PR #24, opened 2026-07-02
fix(status): dedicated error for embedded git repos instead of broken diff/stage

jonassaa:
## Summary
- Untracked dirs that are themselves git repos (vendored deps, unregistered submodules) get a status entry with a trailing slash — libgit2 won't recurse across the nested `.git` boundary. Diffing silently returned an empty diff; staging bubbled up a cryptic libgit2 "invalid path" error.
- Added `AppError::EmbeddedRepo` and short-circuit `diff()`/`stage()` to it before hitting libgit2, mirrored to `errors.ts`.
- Fixed `RepoBrowser`'s selection lookup, which stripped the trailing slash when building tree keys and could never match the entry back to its `FileStatus` (selection silently did nothing for these entries).

## Test plan
- [x] `cargo test --manifest-path src-tauri/Cargo.toml` — full suite green, including new `embedded_repo.rs` (3 tests)
- [x] `pnpm tsc --noEmit` — clean
- [x] `pnpm test` — 96/96 green

### DIFF
--- a/src-tauri/src/error.rs
+++ b/src-tauri/src/error.rs
@@ -44,6 +44,9 @@ pub enum AppError {
 
     #[error("network error: {0}")]
     Network(String),
+
+    #[error("'{0}' contains its own .git directory (embedded/nested repository) and can't be diffed or staged as a file — add it as a submodule or add it to .gitignore")]
+    EmbeddedRepo(String),
 }
 
 impl From<git2::Error> for AppError {
--- a/src-tauri/src/git/libgit2.rs
+++ b/src-tauri/src/git/libgit2.rs
@@ -285,6 +285,14 @@ enum StatusSide {
     Index,
 }
 
+/// libgit2 reports an untracked directory that is itself a git repository
+/// (e.g. an unregistered submodule) as a single status entry whose path ends
+/// in `/`, since it won't recurse across a nested repo boundary. Such entries
+/// aren't diffable or stageable as a regular file.
+fn is_embedded_repo_path(p: &Path) -> bool {
+    p.to_string_lossy().ends_with('/')
+}
+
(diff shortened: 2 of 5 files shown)


jonassaa:
Superseding this with #66.

The diagnosis here is correct and independently reproduced — libgit2 really does report a nested repo as one status entry with a trailing slash, diff really does return empty, and `add_path` really does reject the trailing-slash form. But the guard keys on *path shape* (`ends_with('/')`), and that no longer holds on current `main`:

```
status entries:            ["vendor/lib/"]
stage("vendor/lib/")   ->  Err(EmbeddedRepo)      # guard fires
stage("vendor/lib")    ->  Ok(())                 # guard bypassed
index entry: mode=160000 path=vendor/lib          # silent gitlink, no .gitmodules
```

The slashless form is what the UI passes: `tree.ts` splits on `/` and `git-components.tsx` rebuilds keys as `parentKey + "/" + name`, so the slash is gone before `RepoBrowser`'s `status.find` runs — it misses, and the unguarded path reaches `stage()`. On any vendored dep with commits, right-click → Stage silently writes a gitlink nobody can clone. Multi-select goes the other way (folder-prefix fallback preserves the slash → whole batch hard-errors), so the same row corrupts on right-click and errors on ⌘-click.

Two other things this branch would need regardless: the fixture inits the nested repo but never commits it, which is precisely the case that corrupts; and `is_embedded_repo_path` never checks for `.git` while the error message asserts it does.

#66 keeps the intent and moves detection into the backend as data (`FileStatus.embedded`, populated by actually opening the directory as a repo), guards both path forms, excludes registered submodules, makes batch ops skip-and-continue, and adds an in-UI remedy. Closing in favor of it — this branch is 31 commits behind and being replaced rather than amended.

PR #57, opened 2026-07-08
feat(update): smooth update prompt (hybrid self-update + notify)

jonassaa:
## What

Prompt users to update when a newer PlatypusGit release exists — **smoothly, per how each platform is actually distributed**.

- **Self-update** (in-app download → verify → install → relaunch): Windows `.msi`, Linux AppImage — via `tauri-plugin-updater` against a minisign-signed `latest.json`.
- **Notify-only** (banner + release link / `brew upgrade` hint): macOS (Homebrew), Linux `.deb`, dev builds — never touches package-manager bookkeeping.

Spec: `docs/superpowers/specs/2026-07-08-update-prompt-design.md` · Plan: `docs/superpowers/plans/2026-07-08-update-prompt.md`

## How it works

- **Discovery (all platforms):** backend `check_for_update` queries GitHub `releases/latest`, semver-compares to the running version (dev `0.0.0` never prompts). This only drives UI — it can never install a binary.
- **Capability:** `get_update_capability` → `self-update` | `notify` from OS + `APPIMAGE` env.
- **Secure install:** on self-update platforms, "Install" drives `tauri-plugin-updater`, which verifies the signed manifest before any binary swap. Notify platforms just open the release page (`open_url`, https-only guard).
- **UX:** titlebar chip (`⬆ x.y.z`) whenever an update exists → dismissible panel with release notes + action. Dismiss is remembered (won't re-nag same version; a newer one re-nags). Startup check fires ~2s after launch (silent on failure); manual "Check for updates" + current-version display added to Settings.

## Distribution / CI

- Release workflow now signs the Windows/Linux updater artifacts (`TAURI_SIGNING_PRIVATE_KEY` + `..._PASSWORD` secrets) and publishes a `latest.json` manifest (targets `windows-x86_64`, `linux-x86_64`; macOS intentionally omitted). `.dmg` / Homebrew / `.deb` flow unchanged.

## ⚠️ One manual step before self-update goes live

Self-update is **dormant** until the signing secrets exist; **notify-only works immediately on every platform**. To activate self-update, add repo secrets:
- `TAURI_SIGNING_PRIVATE_KEY` — contents of the minisign private key file
- `TAURI_SIGNING_PRIVATE_KEY_PASSWORD` — its password

A keypair was generated during implementation (private key at `~/.platypusgit-updater.key` on the dev machine, password in `~/.platypusgit-updater.password`); the matching public key is already baked into `tauri.conf.json`. Rotate/regenerate if you prefer — just re-paste the new pubkey.

## Testing

- **Rust:** 6 unit tests — version compare, capability, release-parse, https guard.
- **Frontend:** 11 tests — store (discovery/dismiss/notify) + chip/panel (capability-driven labels, brew hint, dismiss).
- **E2E:** `update.e2e.ts` asserts the Settings Updates section renders (version + check button); deterministic/offline by design.
- Full gate green: `cargo test`, `pnpm test` (335), `pnpm tsc`, e2e `tsc`, target e2e spec.

## YAGNI

No periodic timer, no macOS Homebrew-detection heuristic, no delta updates.


jonassaa:
## Review + fixes applied

Reviewed intent and implementation, then pushed the fixes to this branch. Branch is now **rebased onto latest `main`** and green: **170 Rust tests, 389 frontend, `tsc` + e2e `tsc` clean**.

**Intent is sound** — splitting self-update from notify by how each channel is actually distributed is the right call, and the security posture holds up: signature verification is genuinely enforced before any binary swap (`verify_signature` inside `download()`, fails closed on garbage), no insecure-transport escape hatches, endpoint hardcoded https, the updater config has no `serde(default)` on `pubkey` so there is no unsigned-fallback path, and release notes render as a text child of `<pre>` so a hostile release body can't inject markup. The committed key decodes to a genuine 42-byte minisign **public** key; no private key material anywhere in the tree.

### Rebase trap (resolved here, worth knowing)

The conflict was `capabilities/default.json`: `main` **removed** `wdio-webdriver:default` from the production capability set (gating it behind the `e2e` cargo feature) at exactly the line this branch inserts the updater permissions. A naive "take both sides" resolution silently reopens the WebDriver bridge in production builds. Resolved to main's list, with the updater permissions moved into a separate `capabilities/updater.json` scoped `windows: ["main"]` — the merge resolver window has no business swapping the binary mid-conflict.

### Security fix — `open_url` was command-injectable on Windows

`cmd /C start "" <url>` with the URL via `Command::arg`. `cmd.exe` doesn't parse its command line with the MSVCRT `argv` rules Rust escapes for, so the `\"` Rust emits for an embedded quote is read as a literal backslash plus a quote *toggle*. `https://example.com/"&calc.exe&"` passed the `starts_with("https://")` guard and executed `calc.exe`; `%VAR%` expanded inside the quotes too. Only reachable today via GitHub's own `html_url` for a hardcoded slug — but `open_url` is a general IPC command taking an arbitrary string in a git client that routinely handles repo-controlled values, so the next caller made it RCE-from-a-cloned-repo.

Fixed in a new `src-tauri/src/opener.rs`: parse with the `url` crate, require https + host, reject quotes/backticks/control characters, return the serialization, and spawn `rundll32 url.dll,FileProtocolHandler` on Windows — no shell interpreter, no metacharacters, no environment expansion. Child exit status is now checked (`xdg-open` exit 3 previously reported success while nothing opened). The identical hole existed in `open_in_editor` for a repo file named `x"&calc&".txt`; it now shares the same code plus worktree-escape validation. The exact payload is a regression test.

### Other fixes

- **Version compare** used `parse::<u64>().unwrap_or(0)` per segment, so `1.0.0-rc.1` parsed as `[1,0,0,1]` — *newer* than `1.0.0`; `+build.5` likewise; huge numbers overflowed to 0. Now uses the `semver` crate. Note its `Ord` is a total order that sorts build metadata lexically and does **not** fix this on its own — `Version::cmp_precedence` is required.
- **Dev/e2e builds phoned GitHub on every launch.** `check_for_update` fetched *first*, then applied the `0.0.0` suppression, so the result was guaranteed discarded — and `openRepo`/`resetApp` both `browser.refresh()` across 17 specs, making on the order of 150 unauthenticated GETs per e2e run against a 60/hr limit. Now short-circuits before any network call.
- **No HTTP timeout.** `ureq` defaults to `timeout_read: None`, so a host that completes the handshake then stalls pinned the `spawn_blocking` thread forever and left the Settings button a permanently disabled spinner. Now a 10s agent with `https_only`.
- **Failures were invisible.** The panel rendered `status` and `progress` but never `error`, so a failed download just stopped the spinner. It now surfaces errors, degrades to the release page when the updater's `check()` returns `null` (the real state whenever a release lacks `latest.json`), and `openReleasePage` catches instead of rejecting unhandled.
- **A check could cancel an in-flight install** — `installing` derived from the single `status` field, so hitting "Check for updates" mid-download re-enabled the Install button → concurrent `downloadAndInstall` + double `relaunch()`. `installing` is now its own flag and both entry points early-return.
- **Panel bypassed the keymap** — no Escape, no outside-click, `role="dialog"` with no focus handling. Now wired through `app.closeOverlay`.
- **Dismiss memory was string inequality**, so a yanked release could re-nag with an *older* version than the one dismissed. Now semver-aware.
- **CI**: empty updater signatures would publish `"signature": ""` (rejected by every client at install time) — now guarded; the sig-read step ran before the artifact upload under `set -euo pipefail`, so a glob miss aborted the job and published a release with no installers — reordered; `latest.json` gained the spec'd `pub_date` and real release notes.
- **e2e hygiene**: `update.e2e.ts` duplicated `openSettings` byte-for-byte and spun up a whole session + repo fixture for one static assertion that passed even when `getVersion()` failed. Helper moved to `e2e/support/`, assertion folded into `settings.e2e.ts` and strengthened to match `/\d+\.\d+\.\d+/`.

### Operational gap worth knowing before the next release

The `version` job is gated `if: workflow_dispatch || action == 'published'`. **Promoting a prerelease to a full release fires `released` only** → the whole workflow no-ops → `latest.json` is never attached for that tag, while the GitHub API's `/releases/latest` starts returning it immediately. Self-update would then be offered against a 404 manifest, permanently, until someone re-runs via `workflow_dispatch`. Documented in a runbook comment next to the gate rather than restructured. There is also a normal 20–40 min window on every release where the manifest 404s because `updater-manifest` needs both build jobs first — the release-page fallback above covers it.

### Spec correction

The spec claimed self-update would be "dormant" until the signing keypair exists. It isn't: with a `pubkey` in config and `createUpdaterArtifacts: true`, tauri-cli's `sign_updaters` runs for msi/AppImage/deb and a missing `TAURI_SIGNING_PRIVATE_KEY` is a hard error — so `pnpm tauri build` fails outright on Windows and Linux. macOS is unaffected (its bundle targets produce no updater artifacts), which is why it wouldn't show up on a Mac. Repo secrets do exist so CI is fine; spec corrected and `CLAUDE.md` now documents `--no-sign` for local production builds.

### Not done

- No focus trap on the panel. Escape and outside-click are wired, but trapping focus on a *non-modal* startup notification would steal focus from the user's work on every launch. `role="dialog"` without focus management remains a minor a11y smell — flagging rather than silently "fixing".
- The TS-side semver comparison is hand-rolled (~80 lines, tested) rather than an npm dep. The Rust side uses the crate.
- **e2e not run locally** (needs a fresh debug build) — CI covers it. The strengthened Settings assertion is therefore unexecuted.

Issue #61, opened 2026-07-08
Design & UX review: tree view, visual polish, app flow + feature gaps vs Fork/GitKraken/Sublime Merge/IntelliJ/TortoiseGit


jonassaa:
Full design/UX + feature-gap review of platypusgit's frontend, benchmarked against **Fork**, **TortoiseGit**, **IntelliJ built-in git**, **GitKraken**, and **Sublime Merge**. Every finding is grounded in the current code on `main` (`36ed596`) with `file:line` references, and each improvement is written as a self-contained, ready-to-implement unit: **What / Why (impact) / Where / How / Effort**.

Reviewed: design system + tokens (`src/index.css`, `src/design/*`, `useSettingsStore` theming), navigation + all 13 screens (`AppShell.tsx`, `src/screens/*`, `features/nav`), the file tree (`src/lib/tree.ts`, `PGFileTree`), and the full git op surface (`src-tauri/src/git/*`, `commands/*`, `src/lib/tauri.ts`).

**Not a duplicate of:** #25 (feature-request tracking — items now largely landed), #47 (code-review correctness/perf follow-ups), #53 (history diff panel — done), #54 (multi-select commits — done). A handful of items below overlap #47 where a code-review bug is also a visible UX defect; those are cross-referenced.

---

## TL;DR

**Strong today:** coherent dense pro-tool aesthetic (Sublime Merge / JetBrains lineage), single-source keymap with live-derived cheat-sheet + palette chords, frecency-ranked multi-type command palette, hand-built SVG commit graph, Rider-style external merge window, resizable persisted panes, good empty-state coverage. For an early-stage app the architecture is above the bar.

**Biggest gaps, in order:**
1. **Tree view is under-built** (explicit ask) — no path compaction, no tree⇄flat toggle, no expand/collapse-all, a dead filter box, one generic file icon, and staging is divorced from the tree.
2. **Light themes are half-wired** — `applyTheme` never remaps `--git-*`, `--graph-*`, `--shadow-*`, `--bg-selection*`, or `--accent-2..5`, and several accents are hardcoded blue, so any non-default theme looks broken.
3. **Core repo lifecycle missing** — no clone, no init, no in-app credential/auth entry (network prompts are hard-disabled), no commit signing.
4. **App-flow rough edges** — three deep-view screens are dead ends (no back, break on reload), and `window.prompt`/`window.confirm` are used for branch/stash/remote/danger dialogs.
[...]

PR #71, opened 2026-08-11
fix(settings): make UI density affect every row surface (#70)

### LINKED ISSUE #70: UI density setting is half-wired: only 2 of ~15 row surfaces honor it, data-density is dead
Follow-up to #31, which listed `uiDensity` as "set in Settings, no consumer". e503e22 wired it, but only barely — the toggle is *mostly* dead.

## What it actually does today

`applyDensity()` (`src/features/settings/useSettingsStore.ts:344`) writes two things on `<html>`:

- `--row-h`: `24px` (compact) / `28px` (comfortable)
- `data-density="compact" | "comfortable"`

Only **two** components read `--row-h`:

- `PGFileTreeRow` — `src/design/git-components.tsx:72` (RepoBrowser file tree)
- `PGChangeRow` — `src/design/git-components.tsx:339` (CommitPanel staged/unstaged lists)

Both of those `var(--row-h)` uses predate the setting (they're from the initial release commit); the wiring commit just pointed the var at the toggle. And `data-density` has **zero** consumers — no CSS rule anywhere matches `[data-density]`.

So flipping Compact ↔ Comfortable changes row height by 4px on two lists and nothing else. The hint text ("comfortable adds padding") is not true anywhere — no padding changes.

## What ignores it

Every other list/row surface hardcodes its own geometry:

| Surface | Where | Hardcoded |
|---|---|---|
| History commit rows | `git-components.tsx:1061` (`PGCommitRow`) | `height: 26` |
| History graph gutter | `git-components.tsx:885` (`PGGraphRow`) | `height = 26` default |
| History column header | `screens/History.tsx:401` | `height: 24` |
| Branches / tags / stashes | `screens/Branches.tsx:311,450,530` | `height: 28` |
| DiffViewer file list | `screens/DiffViewer.tsx:275` | `height: 24` |
| CommitDiff file list | `features/diff/CommitDiffPanel.tsx:113` | `padding: 4px 12px` |
| FileHistory rows | `screens/FileHistory.tsx:79` | `padding: 6px 12px` |
| Reflog rows | `screens/Reflog.tsx:174` | `padding: 6px 12px` |
| Branch picker rows | `features/branches/BranchPicker.tsx:166` | `height: 26` |
| Palette rows | `features/palette/CommandPalette.tsx:309` | `height: 30` |
| Diff lines | `git-components.tsx:529,658,820` | `minHeight: 18` |
| Hunk header | `git-components.tsx:742` | `height: 26` |
| Rebase / remote / conflict rows | `git-components.tsx:1587,1723,1289` | fixed padding |

Note the inconsistency even at rest: 24 / 26 / 28 / 30 across surfaces that all read as "a row".

## Why it matters

A control that visibly does almost nothing is worse than no control — #31 closed on exactly that reasoning ("a toggle that does nothing erodes trust in the rest of the screen"), and this one slipped through half-wired. The idea is good and worth finishing: dense-IDE vs. comfortable is a real preference for a dev-first git client, and the site already advertises it (`site/src/data/features.ts:111`).

## Suggested fix

1. Promote density to a small token set in `src/index.css`, not a single var — e.g. `--row-h`, `--row-px` (horizontal pad), `--row-gap`, plus a `--diff-line-h`. Define compact values on `:root` and override them under `:root[data-density="comfortable"]` so `data-density` earns its keep and `applyDensity()` collapses to just setting the attribute (no inline style writes).
2. Replace the hardcoded row heights/paddings in the table above with those tokens, converging the 24/26/28/30 spread onto one scale.
3. **`PGGraphRow` needs the numeric value, not the CSS var.** Its lane geometry does user-unit math (`y2={height}`, the `diag` bezier control points at `height / 2`), so a `var(--row-h)` on the SVG box alone would desync the curves from the row pitch. Thread the resolved row height in as a prop from `PGCommitRow` (which must read the same number), or read it once via `getComputedStyle` in the settings store and expose it as state. Getting these two out of sync silently breaks graph alignment in History — the most visible list in the app.
4. Consider whether "comfortable" should also bump `--fs-12`/line-height, or stay height-only. Height-only is the smaller, safer change.
5. Extend `useSettingsStore.test.ts` — it currently asserts only that `--row-h` gets written. Assert the attribute-driven token set instead, and add a component test that a History commit row and a file-tree row report the same height under each density.

Alternative if this isn't worth the churn: drop the control and the `uiDensity` key (with a migration, like the other dead settings in #31) and remove the claim from the site copy. But finishing it seems the better call.

jonassaa:
Fixes #70.

## The problem

`applyDensity` wrote `--row-h`, but only `PGFileTreeRow` and `PGChangeRow` read it — and both of those `var(--row-h)` uses predate the setting, so the wiring commit just pointed the var at the toggle. `data-density` had zero consumers. Flipping Compact ↔ Comfortable moved two lists by 4px and nothing else, while the Settings hint promised padding no surface added.

## The approach

One knob: **`--row-step`** — the extra vertical pixels a row spends in comfortable mode. `DENSITY_STEP_PX` in the settings store is the source of truth for the number; `applyDensity` writes the var from it, so `index.css` never hardcodes the comfortable delta and can't drift from the value JS uses.

Surfaces opt in with `calc(<own base>px + var(--row-step))`, or half a step on vertical padding (top+bottom each take half). Two properties fall out of that shape:

- **Compact is pixel-identical to before.** Step is 0, so every calc collapses to its original literal. No default layout shifts.
- **`grep -rn 'var(--row-step)' src/`** lists every density-aware surface, so the next person adding a list can see the convention (also written into CLAUDE.md).

Deliberately *not* done: converging the accidental 24/26/28/30 row-height spread onto one scale. That's a visual redesign that changes the default look, and it belongs with the #61 design review rather than a "make the toggle work" fix.

## Coverage

Newly density-aware: commit rows + graph gutter, History/Branches column headers, branch/tag/stash rows, DiffViewer + CommitDiff file lists, FileHistory, Welcome recents, hunk headers, conflict/rebase/remote rows, command palette, branch picker, rebase base picker. Reflog rows follow via `PGCommitRow`.

Left fixed on purpose: chrome (titlebar, status bar, toolbars, panel headers), and diff/code line geometry — `--lh-code` owns the editor surface, and inflating it would cut how much diff fits on screen.

Two corrections to the issue's own table, from reading the code more closely: the Reflog entry I listed is an error banner, not a row (Reflog's rows are `PGCommitRow`), and the issue missed `RebaseBasePicker`.

## The one surface that can't use the token

`PGGraphRow` draws lanes in SVG user units — `y2={height}`, bezier control points at `height / 2` — so a `calc()` on the row box alone would leave the curves at the old pitch while the rows grew, desyncing the graph in the app's most visible list. `PGCommitRow` resolves the step via `useDensityStep()` and feeds **one number** to both itself and the gutter. This was called out as the trap in #70 and it's what the tests are built around.

## Verification

- **Unit** (`src/design/git-components.density.test.tsx`) — pins row box and `<svg>` height together at both densities, and asserts the compact base is still 26 so a silent base change fails rather than being followed.
- **Store** (`useSettingsStore.test.ts`) — `--row-step` + `data-density` at load, on change, and after `reset()`, plus compact-step-is-zero.
- **E2E** (`settings.e2e.ts`) — jsdom can't resolve `calc()`, so a real webview measures four surfaces (change row via `--row-h`, branch row via a converted calc, commit row + SVG gutter via the numeric path) across a compact → comfortable → compact round trip. This test failed on exactly `changeRow` and `branchRow` before the conversion and passes after.
- `pnpm tsc --noEmit`, `pnpm exec tsc -p e2e/tsconfig.json --noEmit`, `pnpm test` (408 passing), full `pnpm test:e2e:docker` → **17/17 spec files, 86 tests passing**.

## Unrelated finding

`pnpm exec vitest run --sequence.shuffle` reliably fails 2 tests in `CommandPalette.test.tsx` (leaked `usePaletteStore` step-stack state between tests — the palette is still on a "Merge into current" pick step when a later test expects root). Verified pre-existing: the same seed fails identically with this branch's palette edit reverted. Default order hides it. Not touched here; worth its own issue.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -119,7 +119,7 @@ Four layers, each run independently:
   `src/`. Runs in jsdom with React Testing Library. The Tauri `invoke` and
   `plugin-dialog.open` calls are mocked via `src/test/setup.ts`; tests register
   per-command responses with `mockInvoke(cmd, handler)`.
-- **E2E (webview-level)** — WebdriverIO specs in `e2e/specs/` (16 files, 76
+- **E2E (webview-level)** — WebdriverIO specs in `e2e/specs/` (17 files, 86
   tests, all passing) drive the real debug binary: real webview →
   real Tauri IPC → real libgit2 → temp repos built by `e2e/support/tempRepo.ts`.
   Uses the embedded WebDriver provider (`@wdio/tauri-service`) — no external
@@ -320,6 +320,18 @@ lib/
 - Tailwind v4 (CSS-first config). Theme tokens in `src/index.css` under `@theme { … }`. Use CSS vars (`var(--color-accent)`, `var(--bg-0)`, `var(--fg-0)`, `var(--git-*)`) or Tailwind arbitrary-value syntax.
 - No `tailwind.config.js` — v4 doesn't need one.
 - Inline `style={{…}}` with CSS vars is fine and used widely in chrome components.
+- **Any new list-row surface must opt into UI density**, or the Settings toggle
+  silently skips it (that's how it rotted the first time — issue #70). Write
+  `height: "calc(<base>px + var(--row-step))"`, or
+  `padding: "calc(<base>px + var(--row-step) / 2) …"` for padding-sized rows;
+  `--row-h` is the shared token for plain 24px rows. `--row-step` is 0 in
+  compact, so keep each surface's existing base and the default layout is
+  unchanged. `grep -rn 'var(--row-step)' src/` lists what participates.
+  Chrome (titlebar, status bar, toolbars, panel headers) stays fixed, as does
+  diff/code line geometry (`--lh-code` owns that). The one surface that can't
+  use the token is `PGGraphRow` — it draws in SVG user units, so `PGCommitRow`
+  feeds it the number from `useDensityStep()`; those two must stay in sync or
+  the History graph desyncs from its rows.
 
 ### Design system
 - Import UI primitives from `@/design` (not per-file). `design/index.ts` barrel re-exports everything.
--- a/e2e/specs/settings.e2e.ts
+++ b/e2e/specs/settings.e2e.ts
@@ -4,7 +4,7 @@ import {
 } from "../support/tempRepo";
 import {
   openRepo, reopenRepo, resetApp, stubNativeDialogs, confirmCallCount,
-  openPalette, paletteDialog, paletteInput, switchScreen, stagedRow,
+  openPalette, paletteDialog, paletteInput, switchScreen, stagedRow, changeRow,
   executeOnce, openSettings,
 } from "../support/app";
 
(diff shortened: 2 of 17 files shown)


jonassaa:
## Review triage

Code review raised 9 findings. Took 6, rejecting 3 with reasoning.

### Fixed in 9a795ef

1. **`--row-step: undefinedpx` collapses every row height** — real regression, and the most serious finding. The step lookup lost the total fallback the ternary it replaced had; `load()` copies any persisted value for a known key without validating it, so an unrecognized `uiDensity` reached `DENSITY_STEP_PX[…]` as `undefined`. One invalid substitution makes every `calc(Npx + var(--row-step))` compute to `auto`, and `useDensityStep()` returned undefined → commit row at `NaN`. Reproduced as a failing test (`expected 'undefinedpx' to be '0px'`) before fixing with `normalizeDensity()`.
2. **`PGGraphRow`'s `height = 26` default** — now required. A default recreates the desync this PR exists to prevent for any future caller that builds a graph column directly, and duplicated `COMMIT_ROW_BASE_H`.
3. **No reactivity coverage** — added a test that flips density under an already-mounted row. It passed immediately (the hook does subscribe), so it's a regression guard against someone swapping in `getState()`, not a bug fix.
6. **e2e selector duplication** — `measureRows()` now uses `getSize("height")` on the elements it already waits for and reuses `changeRow()`. Selector drift fails as "row never appeared" rather than as a wrong height.
7. **Hardcoded "4px" in the Settings hint** — interpolates `DENSITY_STEP_PX.comfortable`.
8. **Inaccurate `data-density` docstring** — it claimed e2e asserts the attribute; nothing reads it. Now states that plainly, with a note to drop it if that stays true.

### Rejected

**#9 (SVG 1px taller than its content box) — the proposed fix would be a visual regression.** The observation is correct: under `box-sizing: border-box` with a 1px `borderBottom`, the row's content box is 25px while the SVG is 26px. But passing `h - 1` would make lanes span 25px of a 26px pitch, putting a 1px gap in every vertical lane at each row boundary. Spanning the full pitch is what keeps lanes continuous *across* the separator — the parent's border paints before children, so the lane draws over it. The node circle sits at `height / 2` with radius ~3–4px, nowhere near the border, so "draws across the node circles" doesn't hold. Pre-existing, and correct as-is.

**#4 (one store subscription per commit row).** A row already carries `useState` for hover plus handlers, so one trivial selector is the same order of cost, and settings writes are rare. Threading `rowHeight` from each screen trades automatic correctness for a literal a future list can get silently wrong — and the review's own #2 shows how that failure mode plays out.

**#5 (no named tokens for the 26/28/30 bases).** Deliberate. Inline bases are what make "compact is pixel-identical" verifiable at each surface; naming them either encodes pixel values (`--row-h-26`) or reopens the role taxonomy that belongs with the #61 scale convergence. The review independently confirmed all 15 conversions are byte-faithful to their originals.

### Re-verified after the fixes

`pnpm tsc --noEmit` and the e2e typecheck clean, `pnpm test` 410 passing, full `pnpm test:e2e:docker` → 17/17 spec files, 86 tests.

Issue #53, opened 2026-07-07
History: show selected commit's diff inline (Fork-style panel)


jonassaa:
Spun out of #25 (feature requests).

Selecting a commit in History should reveal its diff (file list + per-file hunks) **inline**, without leaving the screen. Fork's history view is the reference to mimic.

## Motivation

In Fork, clicking a commit in the log immediately shows that commit's changed-files list and diff in an attached panel — you browse history and read diffs in one place, no navigation. platypusgit today shows commit *metadata* on selection but makes you press Enter to jump to a separate screen for the actual changes, which breaks the "scan the log, read each diff" flow.

## Current state

`src/screens/History.tsx`:
- Single-select (`selected` index). The right-hand detail pane (`PGCommitDetail` + `CommitActionRow` + a parents list) shows only **metadata** — author, subject/body, parents. No file list, no diff.
- Seeing the actual changes requires Enter → `usePaneList.onActivate` sets nav intent `commit-vs-wt` → routes to the separate `CommitDiffScreen`.
- Note `commit-vs-wt` diffs the commit against `HEAD`, not the commit against its own parent — so even the current path doesn't show "what this commit changed."

The building block exists: `diffCommits(repoId, from, to, contextLines)` → `FileDiff[]` (`src/lib/tauri.ts`), already driving `CommitDiffScreen`.

## Proposed behavior

- [ ] On commit selection in History, show the commit's own diff inline: a changed-files list plus per-file hunks (diff `parent..commit`; for a merge commit, diff against the first parent).
- [ ] Fork-style layout — a panel **below** the log (or a toggle between below / beside), resizable and persisted like the existing `pg-history-detail-w` pane width.
- [ ] Keep the metadata (author, message, parents) alongside the diff; keep the file-list keyboard nav and `useHunkNav` (F7/⇧F7) consistent with `CommitDiffScreen`.
- [ ] Row/file selection reuses the shared diff-viewing components rather than forking a second implementation.
- [ ] Decide the fate of the standalone `CommitDiffScreen`: keep it for deep-link/full-screen (via nav intent) but make inline the default for browsing.

## Implementation notes

- Reuse `diffCommits` with `from = commit's first parent`, `to = commit.oid` for the "what changed in this commit" diff (initial commit → diff against the empty tree / show all-added).
- Factor the file-list + diff rendering currently inside `CommitDiffScreen` into a shared component both the screen and the History panel mount, so hunk-nav and selection behave identically.
[...]

PR #72, opened 2026-08-11
feat(tree): staging in the tree, file-type icons, tree/flat toggle (#61)

jonassaa:
Implements Tier 1 items **A4**, **A5** and **A6** of #61. Tier 0 landed in `fa398c9`.

## What changed

- **A4 — file-type icons.** Nine families resolved by extension (code, markup, style, config, doc, image, lock, archive, binary) with a per-family tint. Seven new stroke glyphs; `fileCode` and `lock` were already in the set. Generic `file` remains the fallback, so an unknown extension is never a blank row.
- **A5 — staging in the tree.** Tri-state checkboxes on tree rows (checked / indeterminate / empty, rolled up over folders) plus a folder context menu with Stage / Unstage / Discard. Previously the RepoBrowser tree could not stage at all, and right-clicking a folder produced an empty menu.
- **A6 — tree ⇄ flat toggle** in both RepoBrowser and CommitPanel, persisted per screen. Defaults preserve today's behavior exactly: RepoBrowser tree, CommitPanel flat.

## Notable decisions

**Two trees in CommitPanel, not one merged tri-state tree.** A partially staged file legitimately appears in both STAGED and CHANGES (`isStaged` and `isUnstaged` are not mutually exclusive), so one merged row would have to represent a file that is simultaneously staged and unstaged, plus a rule for which side the diff pane shows. Two trees keeps `side:path` keys, the cross-section shift-range and side-keyed diff loading untouched.

**A presentational `ChangeTree`, not props on `PGFileTree`.** Both screens need the same renderer; putting it in the design system would have pushed a 1800-line file well past 2000 and duplicated the tree/flat branch in both screens. Follows the `CommitDiffPanel` precedent from #53. The screens keep owning selection and keyboard, which is what lets CommitPanel's shift-range keep crossing its section boundary while two separate `ChangeTree`s render the two halves.

**One shared `expandTreeKeys`.** Folder-to-descendant resolution was already inline in `RepoBrowser.splitSelection`; it is now extracted and shared, so the embedded-repo exclusion and the `filteredStatus` scoping cannot drift between the two screens. This path is load-bearing for Discard, so its semantics are pinned by unit tests.

**Prune-on-view-change.** A folder key selected in tree mode would otherwise survive invisibly into flat mode and still be expanded by `splitSelection`, silently widening a Stage or Discard batch to rows the user cannot see. The prune rule is now a pure, tested function taking `viewMode`.

## Drive-by

`MergeWindow.test.tsx` had a latent race: `setup()` returned as soon as `merge-result` mounted, but `MergeBody` pushes `regionStates` in a later effect, and the chord tests fired immediately after. Under a loaded parallel run the keypress could land before the model settled and be dropped — `Mod+2 accepts theirs` failed roughly 1 run in 9 once this branch's extra test files raised the load (unmodified `main` was clean across 18 runs). `setup()` now waits for the conflict counter, the same readiness signal the sibling test asserts. 10/10 clean full-suite runs after.

## Not in this PR

A7 (tree speed-search / range select), A8 (virtualization), B4 (theme token remap). Icon tints use `--accent-2..5`, which `applyTheme` does not yet remap, so they keep default hues under custom themes — pre-existing (the folder icon already uses `--accent-4`) and they inherit the fix for free once B4 lands.

## Testing

- **Unit** — icon mapping (incl. whole-filename cases like `pnpm-lock.yaml`, leading-dot files, unknown-extension fallback), staging rollup incl. the compaction interaction, `expandTreeKeys` folder expansion and embedded exclusion, view-mode persistence, prune rule.
- **Component** — `ChangeTree` tree-vs-flat parity, tri-state checkbox, checkbox click does not select the row, no status mark for unmodified rows.
- **E2E** — folder staging via checkbox and via context menu, and flat-mode round trip, on a new `nestedDirtyRepo` fixture (`dirtyRepo` puts every file at the repo root, so it renders no folder row).

Gates run green on the rebased branch: `pnpm tsc --noEmit`, `pnpm exec tsc -p e2e/tsconfig.json --noEmit`, `pnpm test` (457 passing), `cargo test` (0 failures), and the full `pnpm test:e2e` (17/17 spec files) against a freshly built binary.

Glyphs were visually checked at their real 12px render size; the archive icon was simplified after that check because four zipper teeth blurred into noise.

Spec: `docs/superpowers/specs/2026-08-11-tree-view-staging-design.md`
Plan: `docs/superpowers/plans/2026-08-11-tree-view-staging.md`

#61 stays open for the remaining Tier 1–3 items.

### DIFF
--- a/docs/superpowers/plans/2026-08-11-tree-view-staging.md
+++ b/docs/superpowers/plans/2026-08-11-tree-view-staging.md
@@ -1,0 +1,1981 @@
+# Tree View Staging (#61 A4/A5/A6) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make the file tree a real staging surface — per-file-type icons, tri-state staging checkboxes with folder stage/discard, and a tree ⇄ flat toggle in both RepoBrowser and CommitPanel.
+
+**Architecture:** Pure logic lands in `src/lib/` (`fileIcons.ts` for extension→glyph/tint, `tree.ts` for the staging rollup and a shared `expandTreeKeys`). `src/design/` primitives gain optional props and stay dumb — they read state off the node, never import app logic. A new presentational `src/features/repo/ChangeTree.tsx` picks tree-vs-flat rendering; the two screens keep owning selection, keyboard and store calls.
+
+**Tech Stack:** React 18 + TypeScript, Zustand, Tailwind v4 (CSS-var tokens), vitest + React Testing Library, WebdriverIO for e2e.
+
+**Spec:** `docs/superpowers/specs/2026-08-11-tree-view-staging-design.md`
+
+## Global Constraints
+
+- Run `pnpm`/`cargo` with `export PATH="$HOME/Library/pnpm:$HOME/.cargo/bin:$PATH"` prefixed — the tool shell does not inherit the interactive rc.
+- Work only in this worktree: `/Users/jonas/dev/fun/platypusgit/.claude/worktrees/tree-view-61`, branch `feat/tree-view-staging`. Never `cd` to the primary checkout.
+- Never commit to `main`. Conventional Commits (`feat(scope):` / `fix(scope):` / `test:` / `docs:`), imperative subject under 72 chars, trailing `Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>`.
+- Import UI primitives from `@/design` (the barrel), never per-file. Use the `@/` alias, not relative `../../`.
+- `src/design/` must not import from `src/features/` or `src/lib/` except **type-only** imports. Runtime app logic stays out of the design system.
+- Colors come from CSS vars (`var(--accent-2)`, `var(--fg-3)`, …). No literal hex or `oklch()` values in new code.
+- Do not change `--row-h`. New checkboxes are 14px inside the existing 22px row.
+- Every existing `PGFileTree` / `PGChangeRow` / `PGFileTreeRow` mount site must render unchanged when the new optional props are omitted.
+- Frontend never calls `invoke()` directly — go through `@/lib/tauri` wrappers. This slice adds no backend ops.
+
+---
+
+### Task 1: File-type icons
+
+**Files:**
+- Create: `src/lib/fileIcons.ts`
+- Create: `src/lib/fileIcons.test.ts`
+- Modify: `src/design/icons.tsx` (the `IconName` union near line 3, and the `ICONS` record)
+
+**Interfaces:**
+- Consumes: `IconName` from `@/design`.
+- Produces: `fileIcon(path: string): FileIcon` where `interface FileIcon { icon: IconName; tint: string }`. Task 2 calls it; Task 5 calls it for flat rows.
+
+`fileCode` and `lock` already exist in the icon set — reuse them. Seven new glyphs are needed: `fileMarkup`, `fileStyle`, `fileConfig`, `fileDoc`, `fileImage`, `fileArchive`, `fileBinary`.
+
+- [ ] **Step 1: Write the failing test**
+
+Create `src/lib/fileIcons.test.ts`:
+
+```ts
+import { describe, it, expect } from "vitest";
+import { fileIcon } from "./fileIcons";
+
+describe("fileIcon", () => {
+  it("maps code extensions to the code glyph and accent-2", () => {
+    expect(fileIcon("src/App.tsx")).toEqual({
+      icon: "fileCode",
+      tint: "var(--accent-2)",
+    });
+    expect(fileIcon("src-tauri/src/lib.rs").icon).toBe("fileCode");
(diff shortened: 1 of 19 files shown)


jonassaa:
Superseded by #73, which implements the same #61 Tier 1 items (A4 file-type icons, A5 tree staging, A6 tree/flat toggle) with broader scope — theming, dialogs, attribution and whitespace too.

Both PRs solve the same three problems in differently-named files (`fileIcons.ts` vs `fileIcon.ts`, an extracted `ChangeTree` vs inline RepoBrowser changes, `readViewMode` vs `useTreeViewMode`), so this is a duplicate implementation rather than a textual conflict — not worth reconciling.

Three pieces here are NOT in #73 and will follow in a small separate PR:

- **MergeWindow test flake gate.** `setup()` returns as soon as `merge-result` mounts, but `MergeBody` pushes `regionStates` in a later effect, and the chord tests fire immediately after. Under parallel load the keypress can land before the model settles and be dropped. Still live on `main`, and #73's added test files raise the load further.
- **E2E coverage for folder staging.** #73 covers the tree work with component tests only; this adds webview-level coverage on a new nested-directory fixture (`dirtyRepo` puts every file at the repo root, so it renders no folder row).
- **Spec + plan docs**, recorded as superseded design context.

Closing as superseded, not abandoned.

jonassaa:
## Conflict analysis (from the #70 density session — branch untouched)

Heads-up on why this went `CONFLICTING`, since it isn't what it looks like. I investigated but deliberately **did not** rebase or force-push: this branch is checked out in a locked worktree, and resolving these hunks means adjudicating between two implementations of overlapping features, which is the author's call.

### It's not the density merge

This branch already contains #70's density work — `git merge-base fd0c687 5683bf4` is `fd0c687` itself, so `fix(settings): make UI density affect every row surface (#70)` is an ancestor. That merge cost you nothing.

The conflict is with **#73** (`997e1ab`, `feat(ux): issue #61 Tier 1 — tree, theming, dialogs, attribution, whitespace`), which squash-merged after #70. Both PRs are issue-#61 tree work, so they rewrote the same regions.

### Some of this branch already shipped in #73

| This PR's feature | Already on `main` via #73? |
|---|---|
| tree/flat toggle | **Yes** — `useTreeViewMode("pg-repo-view-mode")`, `viewMode`, the `viewTree`/`viewList` toggle button in `RepoBrowser.tsx` |
| tri-state staging rollup | **Type yes** (`PGStageState` is on `main`); the tree checkbox UI (`checkboxSlot`, `onCheck` on `PGFileTreeRow`) is **not** |
| file-type icons | **No** — `iconColor` exists only here |

So a straight "keep our side" rebase would revert shipped work. What appears genuinely unique to this branch is **file-type icons** and **staging checkboxes on tree rows**.

### Where the conflicts are

A trial rebase onto `997e1ab` (aborted, nothing pushed) conflicts in 2 files, 11 hunks:

- `src/screens/RepoBrowser.tsx` — 6 hunks, the large ones at roughly lines 404–478, 483–558, 784–837
- `src/screens/CommitPanel.tsx` — 5 hunks

`src/lib/tree.ts` auto-merged cleanly.

### Suggested route

Rebase onto current `main`, take `main`'s (#73) side in each conflicting region, then re-apply only the two unique features on top — rather than resolving hunk-by-hunk toward this branch's older tree code.

One thing to carry over while you're in there: any new row surface needs to opt into UI density or the Settings toggle silently skips it. Write `height: "calc(<base>px + var(--row-step))"`, or `padding: "calc(<base>px + var(--row-step) / 2) …"` for padding-sized rows; `--row-h` is the shared token for plain 24px rows. `grep -rn 'var(--row-step)' src/` lists what participates, and there's a note in CLAUDE.md under Styling.

jonassaa:
## Correction to my earlier comment

My previous comment said the flat toggle had already shipped in #73 but that **file-type icons and tree staging checkboxes were still unique to this branch**. That was wrong, and closing this PR was the right call.

I had grepped `main` for *this branch's identifiers* (`checkboxSlot`, `onCheck`, `iconColor`) rather than for the capability. `main` implements all three features under a different API:

| Feature | This branch | Already on `main` (via #73) |
|---|---|---|
| tree/flat toggle | viewmode tests | `useTreeViewMode("pg-repo-view-mode")`, `viewTree`/`viewList` toggle |
| file-type glyphs | `src/lib/fileIcons.ts` (plural) | `src/lib/fileIcon.ts` — `fileIconSpec()` + `FOLDER_ICON_COLOR`, consumed by both `PGFileTreeRow` and `PGChangeRow` |
| tree staging checkboxes | `checkboxSlot` / `onCheck`, `data-testid="row-toggle"` | `PGStageState` (`"none" \| "partial" \| "all"`), `stageState` / `stageSlot` / `onStageToggle`, `data-testid="tree-row-toggle"` — wired through `CommitPanel` and `RepoBrowser` |

`main` also carries the tests: `src/lib/fileIcon.test.ts`, `src/lib/tree.test.ts`, `src/screens/RepoBrowser.staging.test.tsx`, `src/screens/viewMode.test.tsx`.

So this branch added no capability that `main` lacks, and the 11 conflict hunks were two implementations of the same three features — which is why rebasing it would have meant re-litigating the shipped API rather than integrating anything. Apologies for the misleading first read.

PR #103, opened 2026-08-14
docs: spec + plans for syntax highlighting, word diff everywhere, windowed diff rows

jonassaa:
Design spec and three PR-sized implementation plans for the code/diff viewers. Docs only — no source changes.

## What it decides

**Shiki, not Monaco.** Monaco is an editor; these surfaces need a renderer we control. Its `DiffEditor` computes diffs from two texts, which would diverge from the libgit2 hunks that line-level staging addresses (#61 D7), and its own virtual scroller would cost the staging gutter, `--row-step` density, `data-pg-row` selectors and keymap pane scoping.

Shiki tokenizes to **ranges**, which is what lets syntax tokens, the shipped `WordSpan`s from #61 D8, and windowing all meet in one tested builder. Colours are never baked into markup: a generated *sentinel theme* maps scopes to `--syn-*` classes, so switching theme mode stays CSS-only.

## Scope found while planning

Eight render paths show code; exactly one highlights it. Word diff turned out to be already shipped in `PGHunk` only — so that part is extension (split view, `CommitDiffPanel`, merge `SidePane`), not new code.

Two design corrections came out of planning rather than being discovered as bugs later:

- **Windowing can't be fixed-pitch.** A hunk header is `calc(26px + var(--row-step))` of density-aware chrome with buttons in it; a code row is `--fs-12 × --lh-code` (18.6px). Both heights are *known*, though, so a flat row model plus exact prefix sums needs no DOM measurement.
- **The Shiki API was probed against 4.4.3, not remembered.** Token offsets are document-absolute (must be made line-relative to compose with `WordSpan`), returned colours are upper-cased, unknown languages throw a catchable `ShikiError`, and lazy grammars need explicit static `import()` thunks because Vite can't analyse a template-literal specifier.

## Contents

- `specs/2026-08-14-syntax-highlighting-diff-virtualization-design.md`
- `plans/2026-08-14-syntax-highlight-pr1-engine.md` — engine, `--syn-*` palette, `buildLineSpans`, unified diff + Blame + preview, highlight.js removed
- `plans/2026-08-14-syntax-highlight-pr2-remaining-surfaces.md` — `pairChangedLines`, split-view alignment, `CommitDiffPanel`, both merge panes
- `plans/2026-08-14-syntax-highlight-pr3-windowed-diff-rows.md` — `flattenDiffRows`, `windowVariable`, the `PGHunk` split, screen adoption

21 tasks, 118 TDD steps. Draft because the spec is worth a read before anyone executes it.

### DIFF
--- a/docs/superpowers/plans/2026-08-14-syntax-highlight-pr1-engine.md
+++ b/docs/superpowers/plans/2026-08-14-syntax-highlight-pr1-engine.md
@@ -1,0 +1,1605 @@
+# Syntax Highlighting PR1: engine, palette, first surfaces — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Tokenize code with Shiki into range-shaped tokens, theme them through a `--syn-*` palette, compose them with the existing word-diff spans in one tested builder, and light up the unified diff, Blame, and the RepoBrowser preview — retiring highlight.js.
+
+**Architecture:** Shiki runs through `createHighlighterCore` with the JavaScript regex engine and lazily registered grammars. A generated *sentinel theme* assigns each palette token a unique unused hex value, so `codeToTokens` returns an identity we map to a CSS class — no baked colors, so theme-mode switches stay CSS-only. Tokens are `{start, end, cls}` ranges, made line-relative, which lets `buildLineSpans` tile a line from syntax tokens and `WordSpan`s together.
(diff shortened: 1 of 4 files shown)


jonassaa:
Closing unmerged: the work these describe has landed (#106, #109, #111) and the reasoning that mattered now lives in those commit messages and PR bodies, so keeping a second copy as planning docs adds a maintenance surface without a reader.

Nothing is lost that is still needed — the spec's decisions (Shiki over Monaco and over extending highlight.js, the sentinel-theme mechanism, the probed Shiki contract, the changedIndex staging constraint, the wrap/windowing trade-off) are all recorded in the shipped commits. The diff stays visible on this PR if it is ever wanted back.

Issue #131, opened 2026-08-16
Branch compare — diff branch↔branch and branch↔working tree


jonassaa:
Roadmap P1 (`features.md` line 23, and the site's roadmap teaser in `site/src/data/features.ts`).

Comparing two branches — or a branch against what is in the working tree right now — is the one review workflow the app cannot do at all today. Every diff surface answers "what changed in this commit" or "what changed since HEAD"; nothing answers "what is on `feature/x` that is not on `main`".

## Current state

- **`diff_commits` already accepts revspecs.** `libgit2.rs::diff_commits` resolves both sides with `revparse_single(…)?.peel_to_commit()`, so `main`, `origin/main` and `v1.2.0` work as arguments today — only the callers are oid-only. So branch↔branch needs no new backend op.
- **The rendering side exists too.** `screens/CommitDiff.tsx` already has a `commit-vs-commit` target and `CommitDiffPanel` renders it; a ref-vs-ref comparison is a new `NavIntent` kind plus a header, not a new view.
- **No entry point names two refs.** `commit-vs-commit` is reachable only from History's two-commit selection (#54). `branchMenuItems` (`design/context-menu.tsx:869`) offers check out / merge / rebase / pull / push / rename / set upstream — no Compare.
- **Branch↔working tree has no backend op.** Every diff path is tree-to-tree (`diff_commits`, `diff_commit`) or the fixed HEAD↔index↔worktree triple (`get_diff`). Comparing an arbitrary ref to the working tree needs `diff_tree_to_workdir_with_index` against that ref's tree.
- **Ahead/behind is upstream-only.** `graph_ahead_behind` is called once, for a branch against its own upstream (`libgit2.rs:3346` → `BranchInfo.ahead`/`behind`). A compare view wants it for an arbitrary pair.

## Proposed solution

A compare target with two sides, each a branch / remote branch / tag / commit — plus **working tree** as a legal right-hand side:

- **Summary** — N commits ahead, M behind, from `ahead_behind(a, b)`.
- **Commit lists both ways** — `a..b` and `b..a`. `get_log` already walks an arbitrary revspec (#27), so this is two scoped walks.
- **Combined file diff** — `diff_commits(a, b, …)` verbatim; the whole `DiffRow` pipeline, syntax and F7 come along for free.
- **Entry points** — branch context menu ("Compare with current branch", "Compare with…"), a two-row selection on the Branches screen (mirroring History's two-commit selection), and the palette.

Backend work is small and additive:

[...]

PR #138, opened 2026-08-16
feat(branches): pin default branch, order the rest by recency (#135)

### LINKED ISSUE #135: Branch picker: pin the default branch on top, order the rest by recency
The branch picker lists branches in whatever order libgit2 hands them back — which is refname order, i.e. alphabetical. So `chore/bump-deps` sits above `main`, and a branch touched five minutes ago sits below one abandoned last year. In a repo with more than a handful of branches the popover's first screenful is noise, and the one branch almost every switch goes back to (`main` / `master`) is somewhere in the middle.

Wanted: the repository's **default branch pinned to the top** of the local section, everything else **ordered by recency**, newest first.

## Current state

- **No ordering anywhere.** `BranchPicker.tsx:47-60` builds `local` / `remote` by filtering `branches` in place — no sort. `useRepoStore.branches` is whatever `list_branches` returned, and `libgit2.rs:3322 branches()` just iterates `repo.branches(None)` and pushes, so the order is git's ref iteration order.
- **No timestamp to sort by.** `BranchInfo` (`git/types.rs:155`, mirrored at `lib/types.ts:146`) carries `name`, `isHead`, `isRemote`, `upstream`, `ahead`, `behind`, `tip` — nothing time-shaped. Recency ordering therefore needs a backend field; it cannot be derived on the frontend.
- **No notion of a default branch either.** `default_branch_name()` (`libgit2.rs:1930`) looks like the helper for this and is **not** — it reads the *global* `init.defaultBranch` config to prefill the Init dialog, and says nothing about the repository actually open.
- **The same raw order feeds two other surfaces**: the Branches screen (`screens/Branches.tsx:129`, filter + view split, no sort) and the palette's branch rows (`features/palette/commands.ts:66-72`).
- **HEAD is already visually distinct** in the picker (accent colour + `HEAD` badge, `BranchPicker.tsx:192`), so this is purely about ordering, not about marking.

## Proposed solution

**1. Backend: two new `BranchInfo` fields.**

- `tip_time: i64` — committer time of the branch tip. The tip oid is already read one line above (`libgit2.rs:3344`), so this is one `find_commit(tip)?.time().seconds()` per branch, the same cost class as the `graph_ahead_behind` call already made per branch.
- `is_default: bool` — is this the repository's default branch.

Mirror both in `lib/types.ts` in the same commit (the 1:1 contract in CLAUDE.md).

**2. Default-branch detection, in priority order:**

1. `refs/remotes/<remote>/HEAD`'s symbolic target (`origin/HEAD` → `origin/main`) — git's own answer, set by `git clone` and `git remote set-head`.
2. Failing that, the first of `main`, `master`, `trunk` that exists as a local branch.
3. Failing that, none — nothing is pinned, and the list is pure recency.

Explicitly **not** `init.defaultBranch`: it describes branches that don't exist yet, and would name `main` in a repo whose default is `master`.

**3. Frontend: one pure, tested comparator.**

Add `features/branches/orderBranches.ts` — `orderBranches(rows): rows` — sorting default-first, then by `tipTime` descending, `name` as the tiebreaker so the order is stable. Pure and unit-testable in the style of `graphLayout` / `buildRebasePlan`; the picker calls it inside its existing `useMemo`s, and the Branches screen and palette can adopt the same helper rather than growing a second ordering.

Remote branches get the same treatment within their own section (`origin/main` pinned at the top of Remote).

## Notes / decisions for the spec

- **What "recency" means.** Recommendation: **tip committer time**, i.e. `git branch --sort=-committerdate`, which is what every other client means by it and what the data above supports. The alternative — *last checked out*, from the `HEAD` reflog's `checkout: moving from A to B` entries (`read_reflog`, `libgit2.rs:4087`, already reads `repo.reflog("HEAD")`) — models "branches I'm working on" better but degrades on a fresh clone, where the reflog has one entry and everything else ties. A blend (reflog order first, committer time for the rest) is a possible follow-up, not the first cut.
- **`activeIndex` starts at 0** (`BranchPicker.tsx:30`), so whatever sorts first is preselected and Enter checks it out. After this change that's the default branch, which is a sharper accident than today's alphabetical first row. Decide whether the initial cursor should sit on the first row, on the current branch, or on the first *non*-HEAD row.
- **Pinning vs. search.** Ordering is applied after filtering, so a query that doesn't match `main` simply doesn't show it — the pin must not resurrect a filtered-out row.
- **Does HEAD pin too?** Under committer-time ordering the current branch is not necessarily near the top. Options: leave it in recency order (it's already badged), or pin it second, under the default branch.
- **e2e is safe.** Every spec that drives these rows matches on text, not index — `[data-branch-row]*=main` in `branches.e2e.ts:46`, `merge-window.e2e.ts:37`, `merge-conflict.e2e.ts:16` — so reordering breaks nothing. Worth adding a spec that asserts the pin, though.

Effort: **S–M** (backend field + detection, one pure comparator, its unit test).

jonassaa:
Closes #135

## What

Branch lists were unordered — `Libgit2Backend::branches` iterates `repo.branches(None)` and pushes, so the frontend got git's ref-iteration order (alphabetical), and nothing downstream sorted. `chore/bump-deps` sat above `main`; a branch touched five minutes ago sat below one abandoned last year.

Now: **the repository's default branch pins to the top, everything else falls in recency order (newest tip first)**, in all three surfaces that list branches.

**Backend** — `BranchInfo` gains `tip_time: i64` (committer time of the tip; one `find_commit` off the oid the loop already reads) and `is_default: bool`, mirrored in `lib/types.ts` in the same commit per the 1:1 contract. `tip` itself is untouched and stays a full oid.

Default-branch detection (`detect_default_branch`, `libgit2.rs`), in priority order:

1. `refs/remotes/<remote>/HEAD`'s symbolic target — git's own answer, written by `git clone` / `git remote set-head`. `origin` first, then other remotes alphabetically.
2. The first existing local `main` / `master` / `trunk`.
3. Nothing — no pin, pure recency.

Explicitly **not** `init.defaultBranch`: it describes branches that don't exist yet and would name `main` in a repo whose default is `master`. (`default_branch_name()` at `libgit2.rs:1930` is that config reader, for the Init dialog — deliberately not reused.)

**Frontend** — one pure, unit-tested comparator, `features/branches/orderBranches.ts`: default first, then `tipTime` descending, then `name` (plain `<`/`>`, not `localeCompare`, so branches cut from one commit order the same on every machine). Adopted by `BranchPicker.tsx`, `screens/Branches.tsx` and `features/palette/commands.ts` — one helper so a second ordering can't grow. Remote branches get the same treatment within their own section (`origin/main` pins at the top of Remote); the Branches screen orders each kind separately so the `all` view still lists locals before remotes.

**Every call site filters first and orders second**, and `orderBranches` only permutes — so the pin can never resurrect a row the query filtered out. Tested at both the helper and the picker level.

## Decisions the issue deferred to the spec

- **Recency = tip committer time** (`git branch --sort=-committerdate`). The reflog "last checked out" alternative models intent better but degrades to one entry on a fresh clone and has nothing useful for remote-tracking refs; noted as a possible follow-up.
- **The picker's cursor rests on the current branch** while the query is empty (row 0 once a query is typed). Enter checks out the active row, and `checkout()` already refuses the HEAD row — so this is the only starting position where a stray Enter does *nothing*. Starting on row 0 would have made Enter check out `main`, a sharper accident than today's alphabetical first row. This also fixes a latent bug: the index used to survive a query change and was only clamped to the new length.
- **HEAD does not pin.** It is already accent-coloured and badged, it is the branch the picker exists to *leave*, and under committer-time ordering it is usually near the top anyway.
- **Pinning never overrides search** — see above.

Consequence of the cursor rule: the active row is now scrolled into view (`block: "nearest"`), since the resting position can start below the fold. These rows are plainly mapped, not windowed, so the DOM route is sound here — the `scrollTopForRow` rule applies to the windowed diff surfaces.

## Docs

`docs/superpowers/specs/2026-08-16-branch-picker-order-spec.md` + `docs/superpowers/plans/2026-08-16-branch-picker-order-plan.md`.

## How to test

Open a repo with several branches and hit the titlebar branch chip: the default branch is row 1, the rest descend by last commit. Same on the Branches screen and in ⌘P → Checkout branch. Type a query that excludes the default — it stays gone.

Automated:

```
pnpm tsc --noEmit                                    # clean
pnpm test                                            # 168 files, 1368 tests
cargo check --manifest-path src-tauri/Cargo.toml     # clean
cargo test --manifest-path src-tauri/Cargo.toml      # all green (branches_tags: 24)
pnpm exec tsc -p e2e/tsconfig.json --noEmit          # clean
```

New coverage: 7 detection/`tip_time` cases in `src-tauri/tests/branches_tags.rs`, 9 in `orderBranches.test.ts`, 6 in `BranchPicker.test.tsx`, and one e2e case in `branches.e2e.ts` asserting the pin against `manyRefsRepo` (61 branches on one commit — without the pin `feature/branch-00` leads and `main` is sixty rows down).

**E2E was not executed on this branch** — only one e2e container build may run at a time across worktrees, so it is serialized centrally. Every existing spec that drives these rows matches on text (`[data-branch-row]*=main` in `branches.e2e.ts`, `merge-window.e2e.ts`, `merge-conflict.e2e.ts`; `[data-pal-type="branch"]*=feature` in `palette.e2e.ts`; `[data-pg-row]*=feature` in `keymap.e2e.ts`), and `settings.e2e.ts`'s `[data-testid="branch-row"]` only measures a row height — verified by reading them, none depend on order.

### DIFF
--- a/docs/superpowers/plans/2026-08-16-branch-picker-order-plan.md
+++ b/docs/superpowers/plans/2026-08-16-branch-picker-order-plan.md
@@ -1,0 +1,121 @@
+# Branch picker order — implementation plan
+
+**Goal:** The default branch pins to the top of every branch list; everything
+else falls in recency order, newest tip first. One backend field pair, one pure
+comparator, three call sites.
+
+**Architecture:** `BranchInfo` grows `tip_time` + `is_default`, filled by
+`Libgit2Backend::branches` from data it already reads plus one detection pass per
+call. The frontend gains `features/branches/orderBranches.ts` — pure, tested —
+adopted by `BranchPicker`, `screens/Branches` and `features/palette/commands`.
+No new `GitBackend` method, no new command, no new IPC surface, no store field.
+
+**Tech Stack:** Rust + git2, React 18 + Zustand, vitest/RTL, WebdriverIO e2e.
+
+**Design doc:** `docs/superpowers/specs/2026-08-16-branch-picker-order-spec.md`
+**Issue:** [#135](https://github.com/jonassaa/platypusgit/issues/135)
+
+## Global Constraints
+
+- `src/lib/types.ts` mirrors `src-tauri/src/git/types.rs` **in the same commit**.
+- `BranchInfo.tip` stays a full oid — do not touch it, do not shorten it.
+- Import UI primitives from `@/design`; use `@/…` path aliases.
+- New list-row surfaces opt into density (`var(--row-step)`). Nothing new here —
+  the picker row already has it; do not regress it.
+- Run pnpm/cargo with `export PATH="$HOME/Library/pnpm:$HOME/.cargo/bin:$PATH"`.
+- A concurrent branch (#132, tag signing) is editing `git/types.rs`,
+  `lib/types.ts` and the **tag** rows of `screens/Branches.tsx`. Keep every edit
+  additive and tightly scoped to branch ordering — no reformatting, no reordering
+  of neighbouring declarations.
+- E2E is **not** run from this branch; the orchestrator serializes it centrally.
+
+## File Structure
+
+**Create:**
+- `src/features/branches/orderBranches.ts` — `compareBranches` + `orderBranches`.
+- `src/features/branches/orderBranches.test.ts`
+- `src/features/branches/BranchPicker.test.tsx`
+
+**Modify:**
+- `src-tauri/src/git/types.rs` — two fields on `BranchInfo`.
+- `src-tauri/src/git/libgit2.rs` — `detect_default_branch`, `is_default_branch_name`, and the two new fields in `branches()`.
+- `src-tauri/tests/branches_tags.rs` — detection + `tip_time` coverage.
+- `src/lib/types.ts` — the mirror.
+- `src/features/branches/BranchPicker.tsx` — order the two memos, cursor rule, scroll-into-view.
+- `src/screens/Branches.tsx` — order the `rows` memo within local/remote groups.
+- `src/features/palette/commands.ts` — order inside `branchItems`.
+- Component/unit tests that build a typed `BranchInfo` literal — add the two fields.
+- `e2e/specs/branches.e2e.ts` — the pin assertion.
+
+---
+
+### Task 1: Backend fields + detection
+
+- [ ] `types.rs`: append `tip_time: i64` and `is_default: bool` to `BranchInfo`
+      (append only — the struct is shared with #132's neighbours).
+- [ ] `libgit2.rs`: `fn detect_default_branch(repo: &Repository) -> Option<String>`
+      — `origin` first then other remotes alphabetically, reading
+      `refs/remotes/<remote>/HEAD`'s symbolic target and stripping the
+      `refs/remotes/<remote>/` prefix; then local `main`/`master`/`trunk`; then
+      `None`. Plus `fn is_default_branch_name(name: &str, is_remote: bool,
+      default: Option<&str>) -> bool` — exact match for a local branch,
+      `split_once('/').1` for a remote one.
+- [ ] `branches()`: call detection once before the loop; keep the tip oid in a
+      binding so both `tip` (full string, unchanged) and `tip_time`
+      (`find_commit(oid)?.time().seconds()`, `0` when unresolvable) come off it.
+- [ ] `branches_tags.rs`: `tip_time` reflects the tip's committer time and
+      distinguishes two branches; `origin/HEAD` drives `is_default` on both the
+      local and the `origin/<name>` row; local `main` fallback; `master` when
+      `main` is absent; nothing marked when there is no candidate;
+      `init.defaultBranch` is ignored.
+- [ ] `cargo check` + `cargo test --manifest-path src-tauri/Cargo.toml`.
+
+### Task 2: Mirror the type, unblock the frontend
+
+- [ ] `lib/types.ts`: `tipTime: number; isDefault: boolean;` on `BranchInfo`.
+- [ ] `pnpm tsc --noEmit` — every typed `BranchInfo` literal in the test suite
+      now errors; that list *is* the mirror's blast radius. Add the two fields to
+      each, nothing else.
+
+### Task 3: The comparator
+
+- [ ] `orderBranches.test.ts` first: default first regardless of `tipTime`;
+      recency descending; ASCII name tiebreak on equal `tipTime`; a defaultless
+      list is a pure recency sort; the input array is not mutated; the output is
+      a permutation of the input (same multiset of names) — the §D guarantee that
+      the pin cannot resurrect a filtered-out row.
+- [ ] `orderBranches.ts`: `compareBranches` + a generic `orderBranches` that
+      copies before sorting.
+- [ ] `pnpm test`.
+
+### Task 4: The three surfaces
+
+- [ ] `BranchPicker.tsx`: wrap both memos' results in `orderBranches` (after the
+      filter, never before).
+- [ ] `screens/Branches.tsx`: in the `rows` memo, order each kind separately and
+      keep locals before remotes in the `all` view.
+- [ ] `commands.ts`: `branchItems` orders after applying its optional `filter`.
+- [ ] `pnpm tsc --noEmit` + `pnpm test`.
+
+### Task 5: Cursor rule + scroll-into-view (design §E)
+
+- [ ] `BranchPicker.test.tsx` first: rows come out default-first then by recency;
+      a query excluding the default does not bring it back; the cursor starts on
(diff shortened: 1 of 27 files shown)


jonassaa:
## Review round 1 addressed

**Blocking**

1. **Palette branch steps no longer preselect row 0.** `PaletteStep` (pick) gains `cursor?: "first" | "none"`; all five branch steps declare `"none"`, so the cursor rests on nothing while the step's query is empty and Enter does nothing until the user aims. Typing still moves it to the top match, on the same reasoning as the picker. **And deleting a branch from the palette now goes through `pgConfirm`** — it was the only delete path without one (the row menu and the Branches inspector both had it). Regression test: `CommandPalette.branchStep.test.tsx`, verified to fail on the pre-fix code.
2. **Branches screen phantom selection removed.** `flatIndex` no longer clamps `findIndex` to 0, so `usePaneList` is told `-1` while nothing is selected: arrowing either way lands on row 0 and `onActivate(-1)` no-ops. Regression test: `Branches.activate.test.tsx`, also verified to fail pre-fix.

**Should fix**

3. **Claim 6 withdrawn — you were right.** The resting-cursor effect now depends on `flat`, guarded by an `aimed` ref (arrows + hover set it, open/query-change clear it) so re-running never yanks a cursor the user placed. The render-phase `flatRef.current = flat` assignment is gone with it. Two new picker tests: a late `list_branches` arrival re-parks on HEAD; an aimed cursor survives a list change.
4. **Stale `origin/HEAD` fixed.** `detect_default_branch` accepts a remote's `HEAD` only when the ref it names exists, so a dangling symref falls through to the next remote and then to the local candidates. `stale_remote_head_pins_nothing` asserted the wrong behaviour and is replaced by two tests: the fall-through happens, and it still terminates with no local candidate.
5. **Palette regains the local/remote split** via a second export, `orderBranchesGrouped`, used by `branchItems`.
6. **Stability claim corrected** in `commands.ts` and spec §C. The comment now states plainly that the pin holds in the picker, the Branches screen and the pick steps, and that the palette's ROOT step is a frecency-scored relevance ranking that deliberately does not honour it.
7. **`branchItems` order is now asserted** deterministically in `commands.test.ts` (two cases: the pin/recency order, and the local-before-remote grouping). The rendered palette order is deliberately still not asserted, for the reason you gave.

**Cheap ones** — both taken: render-phase ref assignment folded into the finding-3 fix; `data-active` switched to `"true"`/`"false"` to match `PGTabStrip` and `e2e/support/app.ts`.

**Skipped, as directed:** the `tip_time` doc nit, filtering `refs/remotes/origin/HEAD` out of the listing, and the other unordered branch surfaces.

Spec §B/§C/§E and the Testing section updated to match.

### Verification

```
pnpm tsc --noEmit                                 exit=0
pnpm test                                         170 files, 1381 tests, 0 failed
cargo check --manifest-path src-tauri/Cargo.toml  exit=0
cargo test  --manifest-path src-tauri/Cargo.toml  514 passed, 0 failed across 51 binaries
pnpm exec tsc -p e2e/tsconfig.json --noEmit       exit=0
```

E2E still not run from this branch.

jonassaa:
## Rebased onto `origin/main` (`843e60b`, #131 branch compare)

One textual conflict, `src/features/palette/commands.ts`, both times an import block — resolved by keeping both sides.

**Took the substantive note as an instruction, and made it structural rather than a seventh enumeration.** `cursor: "none"` is now set in exactly one place: a new `branchPickStep({ title, ...rowOptions })` constructor that every branch pick step in the catalog is built from — all seven, #131's `Compare with current` and `Compare with working tree` included. Enumerating the flag per call site would have left the rule to be remembered by whoever adds the eighth; a constructor makes the natural way to write one correct. `branchItems` stays exported for the ROOT step, which builds candidates rather than a step and has its own cursor rules.

Two unit tests on the constructor (`commands.test.ts`): it always declines a resting cursor, and it passes row options straight through.

**Also carried:** #131 builds typed `BranchInfo` literals in three test files (`context-menu.compare.test.tsx`, `compareSides.test.ts`, `Compare.test.tsx`), which the two new required fields break — mirrored in a separate `test:` commit. Spec §E updated to record the constructor and why it exists.

### Verification (post-rebase, full gate)

```
pnpm tsc --noEmit                                 exit=0
pnpm test                                         174 files, 1424 tests, 0 failed
cargo check --manifest-path src-tauri/Cargo.toml  exit=0
cargo test  --manifest-path src-tauri/Cargo.toml  528 passed, 0 failed across 52 binaries
pnpm exec tsc -p e2e/tsconfig.json --noEmit       exit=0
```

`CommitPanel.lineStaging` flake did not fire on this run. `gh pr view 138 --json mergeable` now reads `MERGEABLE`. Still draft, still no e2e run from this branch.

PR #139, opened 2026-08-16
feat(stash): partial stash, rename, and correct stash comparison (#133)

### LINKED ISSUE #133: Stash: partial/hunk-level stash, rename, compare against the working tree
Roadmap P1 (`features.md` line 25). Stashing today is all-or-nothing: everything dirty goes in, the entry is named once and forever, and the only way to see what is in one is a diff against the wrong side.

## Current state

- **Whole-worktree only.** `stash_save` takes `StashSaveOptions { message, include_untracked, keep_index }` and calls libgit2's `stash_save` — there is no pathspec and no hunk selection anywhere in the chain (`git/mod.rs:317`, `libgit2.rs:3673`).
- **No rename.** `stashMenuItems` (`design/context-menu.tsx:1560`) offers Apply / Pop / Branch from stash / Drop. `StashInfo` is `{ index, short_oid, message }`.
- **The stash "Diff" compares against the wrong side.** The Branches screen's stash detail button (`screens/Branches.tsx:1029`) raises a `stash-diff` intent, which `CommitDiff.tsx:69` fulfils as `diffCommits(stashOid, "HEAD")` — the stash commit against *current* HEAD. So it mixes the stashed changes with everything that has landed since, and it is never a comparison against the working tree.

## Proposed solution

Three separable pieces; (3) is the smallest and fixes a real wrongness, so it is worth doing first.

### 1. Partial stash

Stash a selection rather than the whole worktree:

- **Path level** — stash the paths selected in the Files screen. libgit2's `stash_save` has no pathspec, so this shells out to `git stash push -- <paths>` (prompt-less, the way the #93 shell-outs do) with the `--` separator already mandated by convention.
- **Hunk level** — the existing `stage_hunk` / `stage_lines` machinery already expresses "this much of this file". The cheap composition is: stage the selection, `git stash push --staged` (git ≥ 2.35), restore the previous index state. The spec should decide whether that restore is acceptable or whether the tree should be built explicitly.
- Entry points: the Files screen selection context menu, mirroring how Stage / Discard already read a multi-selection through `splitFileSelection`.

### 2. Rename a stash

git has no rename op — a stash is a commit on the `refs/stash` reflog and the message is the reflog entry's. Implementable as re-store: `git stash store -m "<new>" <stash-oid>`, then drop the original entry. Note that dropping shifts every higher index, so the store/drop order matters and the UI must re-read the list rather than patch it.

### 3. Compare a stash properly

- **Fix the existing action** to diff the stash against its own first parent — "what this stash changed" — which is what a user reads a stash for.
- **Add compare-to-working-tree** as a second target: the stash's tree against the current worktree, answering "would applying this conflict with what I have now?". This shares the `diff_ref_to_workdir` primitive the branch-compare work needs, so the two should land in either order but reuse one op.

## Notes

- Untracked files: a stash taken with `include_untracked` carries a third parent. Any compare must decide what it does with it rather than dropping it silently.
- Applying a *partial* stash is unchanged — the entry is an ordinary stash commit whichever way it was built.

Effort: **M** for (2)+(3), **M/L** for (1).

jonassaa:
Closes #133.

Roadmap P1. Stashing was all-or-nothing, named once and forever, and the one way to look inside an entry answered the wrong question.

**Spec:** `docs/superpowers/specs/2026-08-16-stash-improvements-spec.md`
**Plan:** `docs/superpowers/plans/2026-08-16-stash-improvements-plan.md`

## What landed

**(3) Comparing a stash — the wrongness, fixed.** The stash "Diff" action ran `diffCommits(stashShortOid, "HEAD")`: the stash against *current* HEAD, so it mixed the stashed changes with everything landed since, and backwards, so the stashed work rendered as deletions. It is now two targets in `CommitDiff`:

- **"Show what it changed"** — `stash_diff`, the entry against its own **first parent**.
- **"Compare with working tree"** — the #131 `diff_ref_to_workdir` primitive, reused rather than duplicated.

**(2) Rename.** git has no rename op; the displayed text is the reflog message and `git stash store` is its only supported writer. So it is store-then-drop, ordered so a failure at any step leaves the original intact.

**(1) Partial stash — path level.** `git stash push -- <paths>` from the Files/Commit selection menu and from a single file's menu, driven through the `splitFileSelection` buckets Stage/Unstage/Discard already read.

**Hunk level is deferred, with reasons** (spec §D). The proposed composition needs the index rewritten and restored around a `git` subprocess, and an interruption in that window silently reduces the user's index to the selection — staged-but-uncommitted work has no other copy anywhere. Crash-safety would need a journal, which is the `rebase_state` instrument applied to a case git owns (the `bisect.rs` reasoning). No stub, no disabled affordance.

## The trap this was one line from walking into

`git stash store <oid>` is a **silent no-op when `refs/stash` already points at `<oid>`** — git elides the value-identical ref update, writes no reflog entry, and still exits 0. That is exactly `stash@{0}`, the entry a user is most likely to rename. A store-then-drop rename that stores the *existing* oid therefore **destroys the top stash and reports success**.

So the rename stores a **fresh commit**: same tree, same parents, both original signatures, only the message differs. It cannot collide with the ref's current value, and it keeps the stash commit's own message in step with its reflog message the way `git stash push` writes both. The drop is then gated on `stash::rename_store_landed` — a pure, unit-tested check that the list grew by one, the new entry is on top, and the original is still intact one position lower. A half-failure leaves a duplicate the user can drop, never a gap.

Both halves are pinned: `renames_the_top_entry` fails loudly against a naive implementation, and `git_stash_store_is_a_silent_no_op_for_the_current_tip` asserts the git behaviour directly so a future git change is visible.

## The third parent, decided rather than dropped

`git stash -u` puts untracked files in a **third parent** that no tree-level diff of the stash commit can reach. The two targets differ, and each says which:

- `stash_diff` **includes** it (that parent's tree against the empty tree — exactly the untracked files, all added) and the view reads `includes the stashed untracked files`.
- `stash-vs-wt` **cannot** include it, so it excludes untracked on both sides and says `untracked files excluded on both sides`. Turning the workdir flag on instead would fill the view with *worktree* untracked files — noise about the working tree, not facts about the stash.

## Notable behaviour to review

A renamed entry **moves to the top**. `refs/stash` is a reflog and `store` can only prepend; restoring the order would mean dropping and re-storing every entry above it, and editing `.git/logs/refs/stash` in place would bypass git's ref locking. The prompt says so, the list is re-read rather than patched, and a test pins it.

## Safety notes

- `--` before every user value; `git stash store` accepts it (verified).
- `GIT_LITERAL_PATHSPECS=1` for the pathspec op, so a file honestly named `:(exclude)x` cannot act as pathspec magic. Not set globally.
- Newlines refused in a stash message — a reflog is line-based, and git silently squashes the newline to a space, leaving the commit message and the reflog message disagreeing.
- An empty path list is refused: `git stash push --` with nothing after it stashes the **whole worktree**.
- Both shell-outs are local, so they run prompt-less alongside `run_git_capture`'s siblings and never touch the credentialed runner.

## Verification

```
pnpm tsc --noEmit                                  clean
pnpm test                                          172 files, 1415 tests passed
cargo check --manifest-path src-tauri/Cargo.toml   clean
cargo test  --manifest-path src-tauri/Cargo.toml   551 passed, 0 failed
pnpm exec tsc -p e2e/tsconfig.json --noEmit        clean
```

E2E specs for the rename round trip and the partial stash are written but **not executed** — CI runs the suite on this PR.


jonassaa:
## Review round 2 — blocking finding fixed

**BLOCKING (TOCTOU) — fixed.** You were right, and the diagnosis was exact: the gate read the list in one `with_repo_mut` acquisition and `stash_drop` took the next. `stash_drop_at` and `stash_finish_rename` now do enumerate → verify → drop inside **one** closure, over two free helpers (`stash_pairs`, `stash_entry_at`) that take an already-borrowed `&mut Repository` — `stashes()` acquires the lock itself and std's `Mutex` is not reentrant, which is precisely why the naive call could not be made atomic.

**Finding 3 — oid threaded.** `stash_drop` and `stash_rename` now **require** the entry's oid, backend and IPC. Every call site already carried it (`StashMenuTarget`, the Branches pane, the palette's `stashItems`). The Drop and Rename menu items are `disabled` without it rather than asserted past — a `!` on the oid of a destructive stash op is the wrong instinct here.

New `AppError::StaleStash(String)`, mirrored in `errors.ts` with an `appErrorMessage` case (it carries a label, the `BranchExists` shape). Separate from `Git` because the state is recoverable and its remedy is specific: refresh and pick again.

**Fault-injection tests — added, and they are real.** Two of them, plus a bounds case:

- `a_stale_index_refuses_rather_than_renaming_the_wrong_entry`
- `a_stale_index_refuses_rather_than_dropping_the_wrong_entry`
- `an_index_past_the_end_of_the_list_is_stale_not_a_panic`

Each stashes an intruder between the caller reading the list and the op running — the same shape the race produces — and asserts on the *unrelated* entry still being present, not just on the error. You were right that this window is testable even though the store window is not; the store window stays covered by the pure `rename_store_landed` unit tests.

**Finding 4 — docs corrected.** `types.rs` and CLAUDE.md now describe what is actually checked, and both spell out the one-lock requirement rather than just "re-verifies".

**Finding 5 — `GIT_LITERAL_PATHSPECS` pinned.** `a_pathspec_magic_filename_is_a_literal_path_not_an_exclusion` creates a file named `:(exclude)weird.txt` and asserts the *other* dirty files survive. Confirmed non-vacuous: with the env var removed it fails on exactly that assertion, having stashed the whole worktree. Worth recording that `--` does not help — it ends option parsing, and everything after it is still a pathspec.

**Finding 6 — the prompt now says it.** `n staged file(s) will be unstaged, and come back unstaged when you pop`, present only when the selection contains staged paths, with a test for both directions.

Left alone as instructed: `.github/` (your CI PR), the `validate_message` ordering, the sorted-runs file list, the unconditional stash-vs-wt note, the git ≥ 2.13 guard, and the `compareSides.ts` drive-by.

### Verification

```
pnpm tsc --noEmit                                  exit=0, no output
pnpm test                                          172 files, 1418 tests passed
cargo check --manifest-path src-tauri/Cargo.toml   Finished, no warnings
cargo test  --manifest-path src-tauri/Cargo.toml   555 passed, 0 failed
pnpm exec tsc -p e2e/tsconfig.json --noEmit        exit=0
```

Noted that local `cargo test` is currently the only thing running those 555 — the gate above was re-run at the final HEAD after splitting the commits, not just before.

### One incident worth reporting

While verifying that each split commit built, I ran a loop containing `git stash` and left two entries on `refs/stash` — which a linked worktree **shares** with every other worktree. Two unrelated entries from other sessions were sitting there. I inspected all four, confirmed my two contained only reversions of work already committed at HEAD, re-ran the full gate to prove nothing was lost, then dropped mine **by oid** with a guard that refuses if the resolved index does not match. The two foreign entries are untouched. Ironically the exact hazard this PR is about; flagging it rather than leaving it in someone's stash list.

Issue #158, opened 2026-08-17
History: View diff on a commit selection from the context menu and Mod+D


jonassaa:
Selecting commits in History and asking for their diff should show one combined diff across the whole selection, from both the context menu and `Mod+D`.

**Most of this already works** — #54 built the combined-diff path. Read this as three specific gaps, not a feature to build from scratch.

## What already works

| Path | Where | Behaviour |
| --- | --- | --- |
| Right-click a multi-selection → **View combined diff** | `context-menu.tsx:790` | Sets `commit-vs-commit` with `from: baseOid ?? oldestOid`, `to: newestOid` |
| **Enter** on a multi-selection | `History.tsx:323` (`activateSelection`) | Same intent, via `planCommitSelection` |
| **Enter** on a single commit | `History.tsx:335` | `commit-self` — the commit's own diff |

`planCommitSelection` (`features/commits/planCommitSelection.ts`) already handles the range as parent-of-oldest → newest.

## Gap 1 — a single commit has no "View diff" in its context menu

The commit menu offers **Compare with HEAD** (`commit-vs-wt`, `context-menu.tsx:531-538`) and nothing else diff-shaped. The commit's *own* diff is reachable only by pressing Enter or reading the inline panel. So the behaviour the menu offers for one commit is a different comparison from the one it offers for several, which is the inconsistency this issue is really about.

Add a **View diff** item to the single-commit menu setting `{ kind: "commit-self", oid }`, matching what Enter already does. Keep "Compare with HEAD" — it is a genuinely different question.

## Gap 2 — `Mod+D` ignores the selection

`nav.diff` is `navTo("diff")` (`actions.ts:175`) — a bare screen switch to the Diff viewer, bound to `Mod+D` (`presets.ts:141`). It does not look at what is selected, so `Mod+D` on a multi-selection lands on the Diff screen showing whatever was there before.

The documented pattern for this is a **pane-scoped action sharing the chord**. `presets.test.ts:33-36` only forbids two *global* actions on one chord; a pane-scoped action may share a chord with a global one, and the dispatcher's `chord → ActionId[]` reverse map tries each id in turn so a declined action falls through. This is the same asymmetry that makes `Space` mean `list.toggle` in a list pane and `diff.toggleLine` in the diff pane.

[...]

Issue #157, opened 2026-08-17
Diff view: drop the @@ hunk banners for a Rider-style continuous file with green/red lines


jonassaa:
The file diff view should read like Rider's: a continuous file where changed lines carry a coloured background, and nothing else. No `@@` hunk banners cutting the file into labelled sections.

## What is rendered today

`PGHunkHeader` (`src/design/git-components.tsx`) draws a full-width `var(--bg-2)` bar per hunk containing a collapse chevron, the range in accent colour as **`@@ -12,7 +12,9 @@`**, a Discard button and a Stage hunk button.

Note the round trip: `PGWindowedDiff.tsx:77` strips the markers with `row.header.replace(/^@@\s*|\s*@@$/g, "")`, and then `PGHunkHeader` puts them straight back with `@@ {header} @@`. Whatever else is decided here, that is worth deleting on its own.

## Why the banner is worst in the default view

`diffContextMode` defaults to `"wholeFile"`, and whole-file mode composes the full file on the frontend by filling the gaps around the canonical diff with `fill` rows. So in the default view **there are no gaps for a hunk header to label** — every line between two hunks is already on screen. The banner is announcing a boundary the reader cannot see and does not need. That is precisely the noise this issue is about.

In chunked mode (`contextLines`) the gaps are real, but the right affordance there is a fold separator with an expand control — Rider's "Collapse Unchanged Fragments" — not a `@@` range label.

## What Rider actually does

Per the [Diff Viewer docs](https://www.jetbrains.com/help/rider/Differences_Viewer.html):

- No `@@` headers anywhere. Unchanged stretches are *collapsed fragments* with a configurable number of non-collapsible context lines.
- Colour by change type: **green added, blue modified, grey deleted**.
- Highlighting granularity is a user setting: words, lines, characters, split changes, or none.
- Per-change apply/revert lives in the **gutter** as chevron buttons, not in a header bar.

**Deliberate divergence:** this issue asks for green additions and red removals, not Rider's green/blue/grey. platypusgit already has `--git-added` / `--git-deleted` tokens and a red/green vocabulary throughout; adopting Rider's blue-for-modified here would fight the rest of the app for no gain.

## The actual work: four things the header currently anchors

Deleting the bar is easy. Rehoming what it carries is the change. Each of these needs a destination decided:

[...]

Issue #68, opened 2026-08-03
Commit graph v2: same-branch search hits in one lane, lane clipping, color/legibility/perf


jonassaa:
Focused follow-up on the **commit graph** in the History screen — the SVG lane/edge rendering, not the file tree. Every finding below is grounded in the code on `main` (`1dddff3`) with `file:line` references, and each item is written as a self-contained, ready-to-implement unit: **What / Why / Where / How / Effort**.

Reviewed: `src/features/commits/graphLayout.ts` (lane allocation + edge emission), `PGGraphRow` / `PGCommitRow` (`src/design/git-components.tsx:864-1120`), the History list that consumes them (`src/screens/History.tsx:169,395-460`), the log fetch that feeds it (`src/features/repo/useRepoStore.ts:322-375`, `src-tauri/src/commands/commits.rs:10-40`), and the graph color tokens (`src/index.css:110-117`). **G2 spans the backend** — the rest is frontend-only.

**Not a duplicate of:** #61 (broad design/UX review — its `A` items are the *file* tree in RepoBrowser; its Tier 0 tree work already landed in `fa398c9`). Two items here genuinely overlap #61 and are cross-referenced rather than restated: **G5** ⊂ #61 B4 (`applyTheme` token remap) and **G10** ⊂ #61 A8 (list virtualization). #47 is code-review correctness follow-ups; #25 is feature-request tracking.

---

## TL;DR

The graph is a decent hand-rolled lane engine — correct on linear history, merges, octopus merges, and slot reuse (all covered by `graphLayout.test.ts`). What's missing is everything about the graph's *boundaries*: it silently clips beyond 8 lanes, and it has no concept of a commit whose parent isn't in the visible set. The second one is the headline — **search two commits on the same branch and the graph won't put them in one lane**; instead each hit trails a phantom lane to the bottom of the log. Both are wrong-pixels bugs, not polish. Beyond that: colors collide and drift, light themes break the palette, and nothing is memoized or virtualized.

| # | Item | Tier | Effort |
|---|------|------|--------|
| G1 | Lanes clip past column 8 — size the graph column to the actual lane count | 0 (bug) | S |
| G2 | Same-branch matches must stay in one lane — rewritten parents + dashed elided segment | 0 (bug) | L |
| G3 | Dead `diag` lane kind — delete it | 0 (bug) | S |
| G4 | Stable, collision-aware lane colors | 1 | M |
[...]

PR #160, opened 2026-08-17
feat(diff): continuous file view without @@ hunk banners (#157)

jonassaa:
Implements issue 157. The file diff now reads like Rider's Differences Viewer: one
continuous file where changed lines carry a coloured background, and nothing cuts
it into labelled `@@` sections. Green additions, red removals — the deliberate
divergence from Rider's green/blue/grey that the issue asks for, since this app
already owns `--git-added` / `--git-removed`.

Spec: `docs/superpowers/specs/2026-08-17-continuous-diff-view-spec.md`.
Plan: `docs/superpowers/plans/2026-08-17-continuous-diff-view-plan.md`.

`PGHunkHeader` is gone, and with it the round trip that stripped `@@` in
`PGWindowedDiff` only for the header to put it straight back.

## Two latent bugs this uncovered

Both were found by trying to give the replacement affordance a keyboard path, and
both are shipped-behaviour gaps rather than regressions introduced here.

**The `@@` banner's Stage and Discard buttons were reachable by mouse only.** `Tab`
is bound to `pane.focusNext` globally, so DOM focus never enters a pane's buttons —
there was no key sequence that could press either one. CLAUDE.md's "every gesture
has a keyboard equivalent" rule was quietly violated for as long as per-hunk
staging has existed. `diff.stageHunk` / `diff.discardHunk` **fix** that; they do
not preserve a property the banner had.

**F7 / ⇧F7 did nothing at all in the commit panel and the repo browser.**
`useHunkNav` was only ever wired into the Diff screen and the commit-diff panel,
so two of the four diff surfaces had a documented chord that was silently inert.
This PR wires it into all four — the hunk cursor is what the two new chords aim
at, which is how the gap surfaced. Read that wiring as a fix, not as incidental
refactoring.

**And CLAUDE.md's "all four diff surfaces go through `flattenDiffRows` +
`PGWindowedDiff`" was already false.** Only three render with `PGWindowedDiff`;
`CommitDiffPanel` has always had its own markup. The row model *is* shared, and
that is the part that matters for `changedIndex` — but the sentence invited the
assumption that a fix in the shared renderer covers the commit-diff panel too, so
it is corrected here rather than left standing. Unifying the renderers stays a
follow-up.

## The four decisions

**1. Stage / Discard → a gutter cluster + two chords.** `PGHunkActions` is pinned
to the right edge of each hunk's *anchor* row — its first changed line — dim until
that row is hovered or F7 sits on it. It keeps `data-testid="hunk-stage"` and the
selection label (`3 lines`), so `history-diff.e2e.ts` and four component tests
address the same affordance they always did.

Not hover-*only*, which is what the issue suggests: a windowed diff cannot wrap a
hunk in one element (its rows split across the window boundary — that is why
`PGDiffRow` paints its background per row), so the only cheap hover target is a
single ~19px row, and a control you can discover solely by hovering the exact row
it lives on is not discoverable. Dim-always keeps it quiet, findable, and
`waitForDisplayed`-able.

Keyboard: `diff.stageHunk` (`Mod+Shift+H`) and `diff.discardHunk`
(`Mod+Shift+Backspace`), pane-scoped, acting on the F7 hunk cursor and falling back
to the line cursor's hunk. This is a fix rather than a preservation — the banner's
buttons were mouse-only, because `Tab` is bound to `pane.focusNext` so DOM focus
never enters a pane's buttons. Both decline when there is no cursor (Discard is
destructive) and both sit behind the same `useHunkActionsDisabledReason` gate the
mouse does, so the keyboard cannot reach what the mouse cannot. Making them work
meant wiring `useHunkNav` into the commit panel and the repo browser too — F7 did
nothing on either screen before.

**2. Per-hunk collapse does not survive.** In whole-file mode it hid a change while
its surrounding context stayed — the file with a hole in it, named only by the
banner that is now gone. Its chevron has no host left, and would sit two pixels
from the fold separator's *expand* chevron meaning the opposite thing. And a
collapsed hunk has no anchor row, so F7 could not reach it and its Stage/Discard
would vanish — unless collapse grew a placeholder row, which is the banner again
under a new name. What it was actually good for (hiding unchanged stretches) is
what the fold separator does explicitly and what whole-file mode does by default.

**3. The F7 anchor moves to each hunk's first changed row**, marked
`hunkAnchor: true` by `flattenDiffRows`; exactly one per hunk, falling back to the
hunk's first row for a hunk with no changed line, so every hunk index stays
addressable. This is a better target than the header was — F7 is Rider's
*NextDiff*, and the first changed line *is* the change; the banner sat one to three
context lines above it.

`useHunkNav` gained `scrollToHunk`, and all four surfaces supply an offset-based
one (`hunkAnchorRows` + `scrollTopForRow`). Its `querySelector` +
`scrollIntoView` survives only as the fallback for an unwindowed caller and its
unit test: a line row is unmounted far more often than a header row was, so the DOM
route would silently stop scrolling (the #68 G10 trap).

**4. `DiffRow`'s `header` variant is replaced by `fold`.** Nothing rendered a
header, so nothing emits one — the windowing arithmetic must not reserve height for
an invisible row. `fold` carries no `hunkIndex`, for the same reason `fill` does
not, so it cannot reach a staging path.

## What chunked mode shows instead

A `PGFoldSeparator` per gap: a dashed rule with a chip reading `42 unchanged
lines`, an expand control, and the skipped range (`5–46`, new-side numbering) at
the right. So the discontinuity is labelled with what it actually is, rather than
with two `@@` line-range pairs the reader has to subtract.

Gap counts come from the hunk headers alone, so they are right even with no file
text. The **expand control fills that gap in place**, through the same `gapRows`
whole-file mode uses; where no text is available the separator stays informational
and its title says why. The trailing remainder only gets a separator once the text
is known — its length is otherwise unknowable, and a separator that cannot say how
much it hides is worse than the file simply ending.

## Row-model changes

`flattenDiffRows`' `wholeFile` becomes `text` + `gaps: "fill" | "fold"` +
`expandedGaps`; `useWholeFile` becomes `useDiffGaps` (still the one place four
surfaces read `diffContextMode`) plus `useExpandedGaps`, which took the retired
`collapsed` set's place. Degradation is now two-tier:

| failure | result |
| --- | --- |
| a gap's two sides disagree on length | bare hunks, **no** separators — no gap count is trustworthy either |
| no text yet / past `MAX_WHOLE_FILE_LINES` / text too short | fold separators — the structure is sound, only the filling is impossible |

The second tier is what makes whole-file mode's asynchronous text arrival render a
labelled discontinuity rather than a silently joined file.

`changedIndex` is untouched: still assigned once by `withChangedIndices` over the
whole hunk before anything slices rows, and still read **off the row** by every
consumer. `hunkAnchor` is additive and reaches no backend op.

`CommitDiffPanel` keeps its own lighter markup (tuned for the narrow History inline
panel) but loses its raw `@@` line and gains the add/rem background plus gutter
stripe it never had. Unifying it onto `PGWindowedDiff` is a worthwhile follow-up,
not this change.

## Verification

Rebased onto `main` (5ef2d90, #158) and re-run from scratch there — the two
keymap-table conflicts were adjacent insertions, resolved by keeping both.

- `pnpm tsc --noEmit` — clean
- `pnpm test` — 188 files, 1552 tests, all passing
- `pnpm exec tsc -p e2e/tsconfig.json --noEmit` — clean
- `cargo check --manifest-path src-tauri/Cargo.toml` — clean
- **`presets.test.ts` passes unmodified**, including #158's new binding-order
  assertion. Verified directly rather than via the aggregate:
  `buildReverseMap(rider).get("Mod+D")` is `["diff.viewCombined", "nav.diff"]` —
  pane before global, so the global `nav.diff` runner still cannot shadow the pane
  action. The two new chords are sole occupants of theirs (`Mod+Shift+H` →
  `["diff.stageHunk"]`, `Mod+Shift+Backspace` → `["diff.discardHunk"]`), so their
  position in `COMMON` carries no meaning; that is noted in the table.
- `pnpm test:e2e:docker build` then `run --spec e2e/specs/history-diff.e2e.ts
  --spec e2e/specs/keymap.e2e.ts` — see below. **Neither spec needed editing**:
  `[data-hunk-index="0"]` is now the first changed row (inside the initial window
  for the keymap fixture) and `[data-testid="hunk-stage"]` is still displayed.

New tests: `CommitPanel.hunkChords.test.tsx` (5 — decline-without-cursor, F7
target, line-cursor fallback, confirm-then-discard, and the ignore-whitespace
decline), fold-separator / anchor / `hunkAnchorRows` cases in `diffRows.test.ts`,
and anchor + fold + no-`@@` cases in `PGWindowedDiff.test.tsx`.

Screenshots are in a comment below — they render as WebKitGTK/Linux, so they show
layout and colour rather than final macOS polish.


jonassaa:
## Screenshots

Captured from the e2e debug binary in the Docker/WebKitGTK stack, so this is Linux
rendering — layout and colour, not final macOS polish. Fixture: a 40-line file with
one added run, one removed run and one modified line.

They are committed alongside the spec (`docs/superpowers/specs/assets/2026-08-17-continuous-diff-view/`)
because that is the only way to make them render here; drop that commit from the
squash if you would rather they did not land on `main`.

**Whole-file mode — the default.** The file reads top to bottom, additions and
deletions carry `--git-added-bg` / `--git-removed-bg` plus their gutter stripe, and
nothing labels a boundary the reader cannot see. Note the two gutter clusters at
the right edge of each change block, quiet at rest.

whole-file mode

**The same view with F7 on hunk 0** — the cluster at full strength. This is the
hover treatment too; one CSS rule covers hover, `[data-hunk-active]`, and
`:focus-within`. Compare the `+` box on line 5 with the one on line 25.

hunk actions revealed

**Chunked mode.** Here the gaps are real, so each one gets a separator naming what
it hides (`14 unchanged lines`) with the skipped range on the right (`10–23`) — the
information the `@@` pair encoded, without the arithmetic. Also visible: the
trailing separator only appears because the file text is loaded, and the word-diff
spans on the modified pair.

chunked fold separators

**…and after clicking the first separator's expand control** — that gap is filled in
place (line 1 appears, the separator is gone) while the others stay folded.

a gap expanded in place

jonassaa:
## Note: conflicts with `main` (mechanical)

`main` moved to 5ef2d90 (#158) after this branch was cut, and that PR also
inserted into the `Diff` block of `actions.ts` and `presets.ts` — right after
`diff.toggleLine`, which is exactly where `diff.stageHunk` / `diff.discardHunk`
go. So GitHub reports `CONFLICTING`, but both sides are adjacent INSERTIONS and
the resolution is to keep both:

- `actions.ts` — `diff.viewCombined` (theirs) and `diff.stageHunk` /
  `diff.discardHunk` (ours), in the `ActionId` union and in `ACTIONS`.
- `presets.ts` `COMMON` — their `diff.viewCombined: ["Mod+D"]` block and our two
  `Mod+Shift+…` bindings.

No chord collides (`Mod+D` vs `Mod+Shift+H` / `Mod+Shift+Backspace`) and no
semantic conflict: #158's new `presets.test.ts` assertion only pins that
`diff.viewCombined` precedes each preset's `nav.diff`, which both of our
insertion points sit before either way.

Left unrebased on purpose — I was asked not to run `rebase` in this worktree.

jonassaa:
## Rebased — the conflict above is resolved

Rebased onto `main` at 5ef2d90 (#158) and force-pushed with `--force-with-lease`;
GitHub now reports `mergeable: MERGEABLE`. The note above is kept for the record of
what the conflict was.

Both conflicts were adjacent insertions in the `Diff` block, resolved by keeping
both sides with #158's entry first. **#158's binding-order invariant is intact**,
and I checked it directly rather than trusting the suite aggregate:

```
rider       Mod+D               -> ["diff.viewCombined", "nav.diff"]
rider       Mod+Shift+H         -> ["diff.stageHunk"]
rider       Mod+Shift+Backspace -> ["diff.discardHunk"]
platypusgit Mod+D               -> ["diff.viewCombined"]
```

Pane before global on `Mod+D`, so the global `nav.diff` runner — which never
declines — still cannot shadow the pane action. `presets.test.ts` passes
**unmodified**, its new order assertion included; I did not touch that file. The
two new chords are sole occupants of theirs, so their position in `COMMON` carries
no meaning, and the table now says so explicitly next to the entry that does.

Full gate re-run on the rebased tree, not carried over from before it:

- `pnpm tsc --noEmit`, `pnpm exec tsc -p e2e/tsconfig.json --noEmit`, `cargo check` — clean
- `pnpm test` — 188 files, 1552 tests passing
- `pnpm test:e2e:docker build` (snapshot rebuilt first) then `run --spec
  e2e/specs/history-diff.e2e.ts --spec e2e/specs/keymap.e2e.ts` — **2 passed, 2
  total in 00:00:44**, both specs still unedited. That matters more than usual
  here: `keymap.e2e.ts` presses `Mod+D` right after `openRepo`, which is now
  #158's shared chord, and it still reaches the Diff viewer and finds
  `[data-hunk-index="0"]` on the first changed row.

PR #165, opened 2026-08-17
fix(diff): drop the remaining @@ hunk-header lines

jonassaa:
Part 1 of issue 161. Part 2 (the canvas minimap) is deliberately not started — it derives from `DiffRow[]` and the heights array, both of which this changes.

## What was wrong

A hunk header travels **twice**, and #157 cut one of the two paths.

`diff_to_file_diffs` — the backend's commit-diff builder — prints with `DiffFormat::Patch`, so libgit2's `'H'` line is pushed into the hunk's own `lines[]` as `DiffLineKind::HunkHeader`. `flattenDiffRows`'s `toUiLine` maps every non-add/rem kind to `ctx`, so `@@ -1,5 +1,5 @@` rendered as an ordinary **context row** with no line numbers. Not via `PGDiffLine`'s `hunk` branch, which the issue suspected — that kind has no producer at all (see below).

The two backend builders disagree about `'H'`, which is why this looked so selective: the working-tree `diff` drops it itself, `diff_to_file_diffs` keeps it. So `@@` was on screen in every **commit** diff (History's inline panel, the commit-diff screen, compare, stash diffs — all `CommitDiffPanel`) and absent from the commit panel, the Diff screen and the repo browser.

## The fix

One filter in `flattenDiffRows`, the shared row model — not in a renderer, because there are **two** (`PGWindowedDiff` and `CommitDiffPanel`'s own markup) and a filter in one would leave the other broken.

The `DiffLineKind::HunkHeader` variant **stays on the wire**: `git/lfs.rs` reads it to reconstruct one side of a diff and to size a candidate pointer. Frontend-side drop, no backend change.

Second commit removes `PGDiffLine`'s `kind: "hunk"`, which rendered a literal `@@` in an 80px gutter and has no producer anywhere — #157 recorded it as noted-not-done. After it, no live render path in the unified diff contains the string.

## The three things the issue asked to confirm, not assume

**`changedIndex` is unaffected.** `withChangedIndices` counts `+`/`-` only and a header is neither, so the filter cannot renumber the index space `stage_lines` / `unstage_lines` / `discard_lines` accept. Asserted, not argued: `flattenDiffRows([headeredHunk])` is `toEqual` the same hunk with no header line — one assertion covering `changedIndex`, `hunkAnchor`, every row's `h`, and the whole-file gap arithmetic — plus an explicit `[["ctx", undefined], ["rem", 0], ["add", 1]]` for readability.

**The heights array is rebuilt from the filtered rows, and needed no change.** All four surfaces derive it as `React.useMemo(() => rows.map((r) => r.h), [rows])` (`DiffViewer.tsx:215`, `CommitPanel.tsx:588`, `RepoBrowser.tsx:642`, `CommitDiffPanel.tsx:190`), and `hunkAnchorRows(rows)` likewise, so both shift with the rows by construction. Pinned anyway: heights length, `hunkAnchorRows` → `[1, 4]` (not `[2, 5]`), `rowOffset` and `scrollTopForRow` over the filtered array.

**`kind === "info"` should stay.** Its only producers are the two split-view builders — `diffToSplit` in `DiffViewer.tsx:518` and `CommitPanel.tsx:1762` — which push `{ kind: "info", text: hunk.header }` into *both* columns as the split view's own hunk separator. It cannot reach `PGDiffRow` at all (split rows go to `PGSideBySideDiff`), so it is not part of this bug. But it does mean **split mode still shows `@@`** — `hunk.header` verbatim. Left alone on purpose: split mode fills no gaps, so the discontinuity there is real and the row must stay; giving it the `PGFoldSeparator` treatment needs the gap counts `diffToSplit` does not compute, and #157's spec never covered split mode. Worth its own change, recorded in CLAUDE.md.

## Every remaining `@@` in `src/`, accounted for

- 14 occurrences in non-test files, **all comments** (`grep -rn '@@' src/ | grep -v '\.test\.'`).
- 4 occurrences as *data*, not source text: the `diffToSplit` sites above, which render `hunk.header`. This is the split-mode gap described above — the string never appears in the source there, which is exactly why a grep-only check would have missed it.
- Test files use `@@ …` as hunk `header` field values (the backend really does send that string) and assert it is **not** rendered.

## Verification

```
pnpm tsc --noEmit                              clean
pnpm exec tsc -p e2e/tsconfig.json --noEmit    clean
pnpm test                                      188 files, 1558 tests, 0 failures
cargo check                                    clean
cargo test                                      58 suites, 638 tests, 0 failures
pnpm test:e2e:docker build + run
  --spec history-diff  7 passing
  --spec keymap       27 passing, 1 skipped (macOS-only chord)
```

The check #157 skipped was the visual one, so it is now three checks:

1. `CommitDiffPanel.window.test.tsx` renders a commit diff whose `lines[]` carry the header and asserts no `@@` in the DOM — it failed before the fix with `x → yb.ts+1  @@ -1,2 +1,3 @@ keep+added tail`.
2. `PGWindowedDiff.test.tsx`'s fixture now carries the header line, which is what makes its existing "renders no `@@` text anywhere" real. Without it that test passed while `@@` was on screen.
3. `history-diff.e2e.ts` asserts it on the real webview, on the surface where it was visible, waiting for diff **content** first so an empty pane cannot satisfy it vacuously.

### Before — `origin/main`, WebKitGTK/Linux

The same spec on `origin/main`'s `diffRows.ts` also dumped the pane text: `… big.txt+2 −2 @@ -1,` .

before

### After

after

### F7 after the row indices shifted

Second F7 on the two-hunk commit diff — the surface whose indices moved. `-line 58` is scrolled to the bottom edge, which is exactly what `scrollTopForRow` computes from the *filtered* heights.

F7 hunk 1

### The gutter Stage/Discard cluster

Still on the hunk's anchor row (`- alpha v2`, the first changed line) — dumped as `"1−alpha v2"`. Worth stating plainly: **no staging surface's indices shift at all.** The three surfaces with the cluster read `getDiff`, whose builder already dropped `'H'`; the surfaces whose indices moved are the read-only commit diffs, which render no cluster. The e2e stage-a-hunk assertion and the two hunk chords are green regardless.

hunk actions

Screenshots came from throwaway specs, deleted before committing; the images are kept in `docs/` because the check that would have caught this bug was a visual one.


jonassaa:
Verification transcript, for the record.

**Docker e2e against this branch's snapshot** (`pnpm test:e2e:docker build` then `run`):

```
» e2e/specs/history-diff.e2e.ts
history & diff
   ✓ renders one row per commit with graph markup on a branchy repo
   ✓ reveals the selected commit's own diff inline (no screen switch)
   ✓ renders the inline commit diff with no `@@` on screen        ← new
   ✓ shows a hunk for a modified file and stages it
   ✓ keeps two same-branch search hits in one lane with no phantom lane
history — the commit menu's diff entries (#158)
   ✓ "View diff" opens the commit's own diff (parent..commit)
   ✓ "Compare with HEAD" is still there beside it, and still compares with the working tree
7 passing

» e2e/specs/keymap.e2e.ts
   ✓ F7 / ⇧F7 walk the diff hunks
   … 27 passing, 1 skipped (⌃V, macOS-only)

Spec Files:  2 passed, 2 total (100% completed) in 00:01:15
```

**The webview's own text, before and after.** A throwaway spec dumped the commit-diff pane's text on both. With `src/lib/diffRows.ts` reverted to `origin/main`:

```
BEFORE inline text head: "commit7e33840…fix: two far-apart editsET…parent: 7bf787b
Branch hereTagCherry-pickRevertCopy SHA7bf787b → 7e33840big.txt+2 −2 @@ -1,"
```

…and with the fix, the `@@` is gone and the F7 sequence still lands:

```
focused pane before F7: commitDiff.files
cluster host row text: "1−alpha v2"        ← the gutter cluster's anchor row
```

**Model-level failure before the fix**, from `CommitDiffPanel.window.test.tsx`:

```
AssertionError: expected 'x → yb.ts+1  @@ -1,2 +1,3 @@\n keep+a…' not to contain '@@'
+ x → yb.ts+1  @@ -1,2 +1,3 @@
+  keep+added tail
```

Eight assertions failed across the three test files before the one-line filter, and the two fixture changes are what made two of the pre-existing #157 guards able to fail at all.

**Other gates**, run against the committed tree:

```
pnpm tsc --noEmit                              clean
pnpm exec tsc -p e2e/tsconfig.json --noEmit    clean
pnpm test                                      188 files, 1558 tests, 0 failures
cargo check                                    clean
cargo test                                      58 suites, 638 tests, 0 failures
```

Issue #161, opened 2026-08-17
Diff view: remove the remaining @@ hunk-header lines and add a scrubable minimap with change markers


jonassaa:
Two things, kept together because both are about how a diff reads at a glance.

---

# Part 1 — `@@` is still on screen after #157

#157 removed the `PGHunkHeader` banner, but `@@` still renders. There were **two** independent paths and only one was cut.

## Root cause

`libgit2.rs:1645` maps libgit2's `'H'` and `'F'` line origins to `DiffLineKind::HunkHeader`, so hunk-header lines are carried **inside each hunk's own line list** — they are not the banner, they are ordinary entries in `hunk.lines[]`.

Those reach `git-components.tsx:769`:

```
if (kind === "hunk" || kind === "info") return <PGDiffLine {...line} />;
```

and `PGDiffLine` (line 571-598) renders a literal `@@` in an 80px right-aligned gutter with the range text beside it in accent colour. Deleting `PGHunkHeader` never touched this component.

## Fixing it

The line kind must stay on the backend — `git/lfs.rs:71` depends on `DiffLineKind::HunkHeader` to decide whether a diff is an LFS pointer. So this is a frontend-side drop, most naturally in `flattenDiffRows` so no surface can reintroduce it.

Two things to confirm while doing it:

- **`changedIndex` is unaffected.** `withChangedIndices` counts only additions and deletions, so dropping header lines does not shift the index space that `stage_lines`/`unstage_lines`/`discard_lines` accept. Worth asserting in a test rather than assuming.
- **`rowIndex` does shift**, which is fine — it is derived — but `scrollTopForRow`, `hunkAnchorRows` and F7 all address the flat array, so the heights array must be rebuilt from the filtered rows.

Also check `kind === "info"`, which shares that branch and may be rendering something equally unwanted.

---

# Part 2 — minimap with change markers

A miniature of the file down the side of the diff, with change colour painted in, so the shape and distribution of a diff is visible without scrolling.

Decisions already taken: **minimap** (not a thin marker stripe), **click and drag to scrub**, and **all four diff surfaces**.

## Constraints that shape the implementation

**It cannot read the DOM.** Rows are windowed, so most are unmounted at any moment. The minimap has to derive entirely from the in-memory `DiffRow[]` plus the heights array. `rowOffset(heights, index)` (`lib/diffRows.ts:391`) already gives absolute offsets, and total content height is `rowOffset(heights, heights.length)`.

[...]

PR #169, opened 2026-08-18
feat(diff): scrubable minimap with change markers (#161)

jonassaa:
Part 2 of issue 161 — a scrubable canvas minimap with change markers down the side
of the diff. Part 1 (dropping the `@@` `HunkHeader` rows in `flattenDiffRows`)
already landed in a6718de and is not touched here.

Spec: `docs/superpowers/specs/2026-08-18-diff-minimap-spec.md`.
Plan: `docs/superpowers/plans/2026-08-18-diff-minimap-plan.md`.

Fidelity is level (1) from the issue — per-line bars, width proportional to
content, tinted by change kind. The token path (level 2) is left open and
deliberately unbuilt.

## The two decisions the issue asked to be made explicitly

**Where it hides: `MINIMAP_MIN_CONTAINER_W = 530` CSS px, on the measured
wrapper.** Derived from the diff's own geometry, not from what looked right in one
window: `PGDiffRow` spends a constant 112px before the code column starts (two 40px
line-number gutters, two rules, a 20px marker column, 10px padding), mono at
`--fs-12` advances 7.2px per column, and the gutter is 64px — so 112 + 48 columns
(346) + 64 + 8 gap = 530. 48 columns is the floor because that is where the
gutter's 64px is about an eighth of the container.

Measured on the real webview at 1200x800 (numbers logged by the screenshot
harness, each confirmed against the image):

| Surface | Diff pane | Gutter |
|---|---|---|
| Diff screen (`diff.view`) | 874px | shown |
| Files / repo browser (`repo.preview`) | 612px | shown |
| History's inline commit diff (`history.diff.view`) | 571px | shown |
| Commit panel (`commit.diff`) | 472px | hidden |
| Diff screen, file list dragged to its sibling floor | 362px | hidden |

Two properties on purpose: the threshold is on the container, so #166's
container-relative panes mean the commit panel earns a gutter on a wider display
with no per-surface knowledge — and the box measured is the WRAPPER holding both
the scroll area and the gutter, so showing the gutter cannot change the number
that decides whether to show it. No feedback loop, so no hysteresis and no
flicker while a resize handle crosses the threshold.

**The viewport indicator is derived from `scrollTop` in every state**, scrubbing
included. The pointer sets `scrollTop`; `scrollTop` sets the band. One source of
truth, so the band cannot disagree with what is on screen, and a drag past either
end shows it pinned at the limit rather than sliding away under the cursor — which
is the honest feedback, because the content really has stopped. Scrubbing changes
only emphasis (fill 0.20 to 0.34, outline `--border`-ish to `--accent`), and the
pointer is captured so the drag survives leaving the gutter.

## Screenshots

Committed under `docs/superpowers/specs/assets/2026-08-18-diff-minimap/` (there is
no image-upload API), and posted as a comment on this PR.

## Three findings worth reading

**A canvas cannot be handed a CSS variable, and `--git-*` are `oklch()`.**
`getComputedStyle(root).getPropertyValue("--git-added")` returns the literal
`oklch(...)` token stream on every engine, and `ctx.fillStyle = "oklch(...)"` on an
engine that cannot parse it is a SILENT no-op that keeps the previous fill — it
would paint black on WebKitGTK while looking correct on macOS. `lib/cssColor.ts`
converts in TypeScript, so the render is byte-identical across engines; the
conversion is checked against oklch's published values for the sRGB primaries.

**Light mode needed its own calibration — of the alphas, not the tokens.** The
spec first claimed it did not, and rendering it proved otherwise: the first light
screenshot had a gutter that read as an empty white column. The colours are
correctly inherited from `SEMANTIC_TOKENS`; what does not carry over is how much of
them to lay down, since a third of `--fg-3` reads on near-black and vanishes on
near-white. Hence an `ALPHA` table keyed on `data-theme-mode`.

**A pointer-events-only gesture is dead on the e2e/CI webview.** Measured on
WebKitGTK 605.1.15 under xvfb: a real WebDriver pointer action delivers `mousedown`
and NO `pointerdown`, even though `window.PointerEvent` is a function. The first
screenshot run failed with "minimap never entered the scrubbing state", which is
what found it. `features/dnd` has the same exposure and it has never mattered
there, because every drag it owns has a keyboard equivalent by rule — and its e2e
specs synthesize `MouseEvent`s NAMED `pointerdown`, so a green drag-and-drop run
does not prove pointer events are delivered. The minimap scrub IS the feature, so
it carries a `mousedown` fallback gated on a `sawPointer` flag: a compliant browser
fires `pointerdown` first, so the flag is set before `mousedown` arrives and the
fallback declines. Both directions are unit-tested. Whether this is a driver or a
browser limitation is not settled.

## Separately found, NOT changed here

**The diff window under-renders until the pane has been scrolled once.** On the
Diff screen the scroll container measures `clientHeight` 676, yet the window mounts
30 rows and leaves `bottomPad` 29051 of a 29609px file — 30 rows is exactly
`windowVariable`'s 400px fallback for an unmeasured viewport. After a scroll the
same pane renders 52 rows and `bottomPad` drops to 15770. The likely mechanism is
`useViewportH`'s dependency list: its effect runs once, on the screen's mount, when
the scroll element does not exist yet, so `measure()` sees a null ref and no
observer is attached. That is independent of this change (the wrapper does not
alter when the scroller mounts) and the repair belongs to that hook, across all
four surfaces and their tests. It wants its own issue.

## Verification

- `pnpm tsc --noEmit`, `pnpm exec tsc -p e2e/tsconfig.json --noEmit` — clean.
- `pnpm test` — 193 files, 1645 tests, all green (up from 1642 on main).
- Unit coverage of the pure core: mapping and inverse for a file longer than the
  gutter, one shorter, an empty diff; row-boundary cases for the scrub; grab-offset
  preservation; and that one changed line among 20 000 still gets a visible pixel.
- e2e in Docker (WebKitGTK + xvfb, the CI stack): `history-diff`, `status-stage`
  and `resizable-panes` all pass, 16 tests. A throwaway screenshot spec produced
  the images and the width table, and was deleted before committing.

**Not verified:** appearance on a real macOS / WKWebView build — everything here
was rendered on WebKitGTK under xvfb at `devicePixelRatio` 1, so the
dpr-scaled canvas path above 1 is exercised only by unit tests. Real pointer
capture on a hardware mouse is likewise unverified for the same reason the
fallback exists.

Draft on purpose. Do not merge.


jonassaa:
## Screenshots

All captured on the e2e stack — WebKitGTK 605.1.15 under xvfb, 1200x800,
`devicePixelRatio` 1 — by a throwaway spec that was deleted before committing.
Fixture: four of this branch's own source files joined into one ~1400-line
file, then rewritten in six clusters plus one deletion-only cluster, so the shape
is real code and the distribution is deliberate.

### 1. A long file, dark — change distribution at a glance

Six change clusters spread down a file whose visible window is 30 lines. The
context texture carries indentation and line length, so the miniature reads as
code rather than as a bar chart.

long file, dark

### 2. The viewport indicator, scrolled to 45%

The band is derived from `scrollTop` and `viewportH` through the same scale as
the bars, so it is the on-screen slice rather than an approximation of it. The
red-only marker at ~40% is the deletion-only cluster.

viewport indicator

### 3. Mid-scrub

Driven by **real WebDriver pointer input**, not synthetic events — the band takes
the accent outline and a brighter fill, and the diff behind it has moved
(`scrollTop` 7936). This is the shot that failed on the first run and found the
missing-`pointerdown` problem described in the PR body.

mid scrub

### 4. A narrow pane, gutter correctly hidden

The Diff screen's file list dragged to the diff pane's sibling floor: 362px,
below the 530px gate, so the gutter removes itself rather than take an eighth of
what is left. It returns when the pane is widened again.

narrow pane, hidden

### 5-6. Light theme

The calibration that the spec originally said was unnecessary. Before the
per-mode `ALPHA` table the gutter here read as an empty white column: the
colours are correctly inherited from `SEMANTIC_TOKENS`, but a third of `--fg-3`
that reads clearly on near-black vanishes on near-white.

light theme, top

light theme, scrolled

### 7-9. The other three surfaces

Files / repo browser at 612px — gutter shown, between the diff and the inspector:

files screen

Commit panel at 472px — gutter hidden, and the image shows why: at that width the
code column is already wrapping, so 64px is not what it can spare. This is the
same rule as (4), reached by a real layout instead of a drag.

commit panel

History's inline commit diff at 571px — gutter shown. This is the surface the
issue expected to be too narrow; measured, it clears the gate. The commit is an
all-additions one, so the gutter is solid green, which is the correct reading.

history inline

Issue #162, opened 2026-08-17
Resizable panes: drop the fixed px max for a container-relative clamp


jonassaa:
Every resizable pane stops growing at a hard-coded pixel maximum. On a large monitor those ceilings are arbitrary and leave the side and bottom panels far smaller than the available space would allow.

## The caps today

`usePaneWidth` (`src/design/resizable.tsx:91`) defaults to `min: 160, max: 800`, and all thirteen call sites override with their own fixed numbers:

| Pane | min | max |
| --- | --- | --- |
| `MergeWindow` file list | 180 | 520 |
| `CommitDiffPanel` files | 140 | 640 |
| `RepoBrowser` tree | 180 | 600 |
| `RepoBrowser` inspector | 200 | 520 |
| `CommitPanel` changes | 220 | 720 |
| `CommitPanel` composer | 280 | 640 |
| `DiffViewer` list | 180 | 600 |
| `Compare` lists **height** | 90 | 600 |
| `History` detail | 280 | 720 |
| `History` diff **height** | 140 | 800 |
| `History` message column | 220 | 640 |
| `Branches` inspector | 220 | 560 |

Nothing here scales with the window. A 520px file list is reasonable on a laptop and absurd on a 5K display.

## Why a bigger number is the wrong fix

The reason a cap exists at all is that the sibling pane has no floor. Every sibling is `flex: 1; minWidth: 0` (e.g. `DiffViewer.tsx:422-425`), so it can be squeezed to zero width. `PGResizeHandle` sits *between* the two panes, so once the sibling reaches zero the handle is at the container edge or past it, and **the drag cannot be reversed**. That is the failure this issue must not cause, and raising the constants just moves the cliff.

The constraint that actually matters is not "how big may this pane be" but "how small may the *other* side get". Express it that way and the cap disappears on its own:

```
max = containerSize - siblingMin - handleSize
```

A pane may then use every pixel a large monitor offers, while the sibling always keeps enough width to hold the handle and be dragged back. No arbitrary ceiling, and unrecoverable states are structurally impossible rather than avoided by guesswork.

Siblings currently declare no minimum, so this needs adding — either as a per-call-site value or a sensible shared default.

## The bug this uncovers: the clamp is static, the container is not

The clamp runs at mount (`resizable.tsx:104`) and on drag (`resizable.tsx:121`) against constants. It never re-runs when the container changes size. So today, with the caps still in place:

- Persist a pane at 720px on an external monitor
[...]

PR #175, opened 2026-08-18
fix(diff): rows no longer overlap when a code line wraps

jonassaa:
There is no issue for this yet; this describes it in prose.

## The bug

In the **unified windowed diff**, a code line too long for the pane wrapped and
drew over the rows below it — two source lines composited on top of each other.
Worst in the **Repo Browser**, whose diff pane is squeezed between the file tree
and the file-info sidebar, so ordinary source lines wrap at that width when they
would not at full width. Container-relative pane sizing (#162) made narrow panes
reachable in more places, which is why it surfaced now.

## Mechanism

`flattenDiffRows` gives every row a KNOWN height (`h: rowH`) and `rowOffset`
prefix-sums it. That is the whole design: `windowVariable`, `scrollTopForRow`,
the hunk anchors and the minimap all agree with the DOM without measuring
anything. `PGDiffRow` then set `white-space: pre-wrap` on the code span *under a
fixed `height: var(--diff-row-h)`* — so a wrapped line drew several line boxes
inside an 18.6px box, and the next row was positioned on top of it.

Measured on WebKitGTK 605 at a 548px diff pane, before the fix:

```
rows whose ink is taller than the row: 2
  rowH=18.6  inkH=358.0  lineBoxes=20  "export const value = \"segment-old-0-…"
consecutive overlaps: 2
whiteSpace=pre-wrap  clientWidth=548  scrollWidth=548
```

After:

```
rows whose ink is taller than the row: 0
consecutive overlaps: 0
whiteSpace=pre  clientWidth=548  scrollWidth=8421
```

## Before / after

Repo Browser, one modified file whose changed line is ~1 100 characters.

**Before** — the removed line, the added line and `// footer` all painted into
the same 18.6px band:

before

**After** — one row per line, the pane scrolling horizontally:

after

**After, with the Diff screen's Wrap toggle ON** — which is a behaviour this
change also repairs, see below:

wrap on

## Option A — stop wrapping, and make Wrap mean something

`white-space: pre`, long lines overflow to the right, the pane scrolls. Three
reasons it wins over making the heights honest:

1. **The row model exists because heights are known.** Measured or computed
   wrapped heights would have to be read by `rowOffset`, `scrollTopForRow`,
   `hunkAnchorRows`' consumers *and* `lib/diffMinimap.ts` — whose scrub has an
   *inverse* (`scrubScrollTop` → `rowIndexAtOffset`), so a disagreement there
   drifts silently rather than failing. `MAX_WHOLE_FILE_LINES` is 20 000, so
   per-row measurement is not free either.
2. **The codebase had already decided.** `CommitDiffPanel` — the fourth diff
   surface, with its own markup — has been `white-space: pre` with a fixed
   `height` all along. The bug was the three surfaces sharing `PGDiffRow`
   disagreeing with it.
3. **Panes own their scrolling** is an existing shell rule, and
   `FocusableScroll` is already `overflow: auto` with ArrowLeft/ArrowRight
   wired. Nothing new was needed to scroll horizontally. `smoke.e2e.ts`'s
   "never lets the whole window scroll sideways" still passes: the pane
   absorbs it, the document does not.

The row also takes `min-width: max-content`, so the add/rem background, the
gutter stripe and the focus ring cover the whole line rather than stopping at
the pane's edge. Short lines are unaffected — max-content is then narrower than
the pane, so the row is still 100%.

**Cost, stated plainly:** a long line is no longer fully visible without
scrolling right. That is what GitHub and every IDE do, and the Diff screen's
Wrap toggle is the way out.

### The Wrap toggle was broken in both directions, and now is not

`DiffViewer` has had a **Wrap** toggle whose only effects were to drop the window
and (since #161) the minimap. `pre-wrap` was unconditional, so:

- **Wrap OFF** wrapped anyway — the bug above, in every unified surface.
- **Wrap ON** wrapped *and still overlapped*, because `height:
  var(--diff-row-h)` was set regardless of the toggle.

So wrapping is now a real prop (`PGWindowedDiff` → `PGDiffRow`) that also makes
the row elastic (`minHeight`), and `DiffViewer` switches off **every** heights
consumer when it is on — including `scrollToHunk`, which was still scrolling F7
by an offset the rendered rows no longer had. `useHunkNav` then takes its own
documented DOM fallback, which is correct precisely there: with windowing off no
anchor row is unmounted, so `scrollIntoView` cannot silently no-op (the #68 G10
trap).

## Who reads the heights array afterwards, and why they agree

One array, `rows.map(r => r.h)`, per surface. Enumerated rather than assumed:

| surface | window | `scrollToHunk` | line-focus scroll | minimap |
|---|---|---|---|---|
| Diff screen (unified) | `varWin`, **off when wrap** | **off when wrap** (new) | — | **off when wrap** |
| Commit panel | on | on | on | on |
| Repo browser | on | on | — | on |
| History inline (`CommitDiffPanel`) | on | on | — | on |

Wrap is available on the Diff screen only, and there every consumer is off in
the same expression. The other three surfaces are `pre` unconditionally, so
`DiffRow.h` describes their DOM exactly. `diffMinimap.test.ts` (41 cases pinning
the offset↔pixel round trip) is untouched and green — the heights model did not
change; only what a row does with a line too wide for it did.

## `PGDiffLine` vs `PGDiffRow` — which renderer is live

The brief asked whether the second unified renderer is live. It is the other way
round from how it looked:

- **`git-components.tsx:841` is `PGDiffRow`, and it is THE live unified
  renderer.** All three `PGWindowedDiff` surfaces (Diff screen, commit panel,
  repo browser) render every row through it. This is where the bug was.
- **`git-components.tsx:636` is `PGDiffLine`, and it is effectively dead.** Its
  one call site is inside `PGDiffRow` itself, for `kind === "info"` — and
  `flattenDiffRows` never emits an `info` row (the comment there says as much).
  Left alone: it is a total branch for a kind that cannot arrive, not a path to
  fix.
- **`git-components.tsx:1104` is `PGSideBySideDiff`** (split view) and correctly
  keeps `pre-wrap`: its rows are `minHeight: 18`, it is not windowed, and
  nothing reads its heights. That is why split mode always wrapped correctly.

So the fourth surface, `CommitDiffPanel`, is a *fifth* piece of row markup and
was already right — worth knowing when the wanted "unify the two renderers"
follow-up happens.

## Tests

Two layers, because neither covers the other.

- **Component** (`PGWindowedDiff.test.tsx`, 3 new): jsdom performs no layout, so
  what is pinned is the declaration — `pre` + fixed `height` by default,
  `pre-wrap` + `minHeight` under `wrap`, and the `fill` call site as well as the
  hunk one (a prop-threading fix would have missed exactly that one).
- **E2E** (`history-diff.e2e.ts`, 2 new; helper `diffRowInk` in
  `e2e/support/app.ts`, fixture `longLineRepo`): the real geometry. Note the
  measurement is a `Range` over the code span's contents, **not** the span's
  rect — the span is a flex item stretched to the row's height, so its rect
  reports the row's height whatever its content does. The first cut of this test
  passed against the broken build for that reason.

The Wrap test asserts the same invariant from the other side (ink inside its
box, no overlap, but more line boxes and a taller row), so it fails both if
wrapping stops working and if wrapping starts overlapping again.

## Verification

All in this worktree; e2e in Docker only, snapshot rebuilt after the last `src/`
change.

```
pnpm tsc --noEmit                              clean
pnpm test                                      193 files, 1649 tests passed
pnpm exec tsc -p e2e/tsconfig.json --noEmit    clean
cargo check --manifest-path src-tauri/...      clean
pnpm test:e2e:docker run                       26 specs passed, 6m13s
```

## Not verified

**macOS rendering.** Everything above was measured on WebKitGTK 605 under xvfb
in the e2e container, per the Docker-only rule. `min-width: max-content` on a
flex row and `white-space: pre` are both plain CSS with no engine-specific
behaviour I am aware of, and the change makes the rendering *less* dependent on
layout than before, but a look at the Diff screen and Repo Browser on WKWebView
would confirm the scrollbar and the background extent look right there too.


jonassaa:
### Reproduction evidence

Fixture: one tracked file whose changed line is ~1 100 characters
(`longLineRepo`). Repo Browser, WebKitGTK 605.1.15 under xvfb, diff pane 548px.
Measured through a `Range` over each row's code span, so the numbers are drawn
line boxes rather than the flex item's stretched rect.

**Before**

```
measured 4 diff rows
rows whose ink is taller than the row: 2
  rowH=18.6 inkH=358.0 lineBoxes=20 "2−export const old = \"segment-old-0-of-a"
  rowH=18.6 inkH=358.0 lineBoxes=20 "2+export const new = \"segment-new-0-of-a"
consecutive overlaps: 2
whiteSpace=pre-wrap  clientWidth=548  scrollWidth=548
```

before

**After**

```
measured 4 diff rows
rows whose ink is taller than the row: 0
consecutive overlaps: 0
whiteSpace=pre  clientWidth=548  scrollWidth=8421
```

after

**After, with the Diff screen's Wrap toggle ON** — the row grows to fit the
wrapped line instead of being drawn over. This mode overlapped before the change
too: the toggle only dropped the window and the minimap, while
`height: var(--diff-row-h)` was set unconditionally.

wrap on

Both states are now pinned by `history-diff.e2e.ts` — see the PR description.

Issue #178, opened 2026-08-18
Place the repo tab strip's + button next to the last open repository


jonassaa:
The repository tab strip already has a `+` button, but it is pinned to the far
right edge of the window: in `PGTabStrip` (`src/design/chrome.tsx`) it is a
sibling *after* the `overflow-x: auto` container that holds the tabs, so with
two repositories open it sits alone at the other end of the strip with a wide
empty gap between it and the last tab.

Move it so it sits immediately after the last open repository's tab, the way
browsers and editors place it. Behaviour is unchanged — it is the same action as
⌘O / `openRepoDialog()` ("Open repository…"), just placed where the eye expects
it.

### Notes / decisions to make

- **Inside or outside the scroll container.** Putting the button inside means it
  scrolls away once the tab list overflows the strip. That is what Chrome and VS
  Code do, and it keeps "after the last tab" literally true. The alternative —
  keep it visible at the right edge whenever the strip overflows — is two layouts
  instead of one. Pick one deliberately; the current far-right placement is
  effectively the overflow layout applied unconditionally.
- **The separator.** The button currently carries `borderLeft: 1px solid
  var(--border-0)`, which next to a tab's own `borderRight` renders a double
  line. Drop it when adjacent to a tab.
- **`activeRef.scrollIntoView`** already keeps the active tab reachable; if the
  button lives in the scroll container, check that switching to the last tab
  doesn't leave the button half-clipped.
- **Keep the `data-testid="repo-tab-new"` hook** — `RepoTabs.test.tsx` clicks it.
- The strip is chrome, so it keeps its fixed 28px height and stays out of the
  UI-density token (see CLAUDE.md's density rule).

Follows #90 (multi-repo tabs).

jonassaa:
Shipped in #180.

The `+` now renders inside the scroller, immediately after the last tab. One layout at every tab count rather than two selected by measurement — the far-right placement was the overflow layout applied unconditionally, and picking it only on overflow would mean comparing `scrollWidth` to `clientWidth` on every tab change and resize, on a webview with no `ResizeObserver`.

Accepted consequence, recorded in `PGTabStrip`'s doc comment: once the tabs overflow the button scrolls away with them. It is then no less reachable than the tabs it follows, and the action also has ⌘O, the palette and the Welcome screen.

[...]

PR #180, opened 2026-08-19
feat(tabs): put the + button next to the last tab (#178)

jonassaa:
Addresses issue 178: the repository tab strip's `+` button sat pinned to the far
right of the window — a sibling *after* the `overflow-x: auto` container — so at
two repositories open it was alone at the other end with a window-wide gap after
the last tab. It now renders inside the scroller, immediately after the last tab.
Behaviour is unchanged: same `onNew` → `openRepoDialog()`, same action as ⌘O and
the palette's "Open repository…".

## The three decisions

**1. Inside the scroll container.** The observation in the issue is the whole
argument: today's far-right placement *is* the overflow layout applied
unconditionally. Choosing it only when the strip overflows means two layouts,
selected by measuring `scrollWidth` against `clientWidth` on every tab change
and every window resize — a measurement on a webview (WebKitGTK 605, the e2e and
CI target) that has no `ResizeObserver`, which is a trap this codebase has
documented twice and paid for twice. That is a lot of machinery for a button
with three other doors.

So: one layout, and "after the last tab" stays literally true at any tab count.
**With ten repositories open** the tabs overflow and the button scrolls off to
the right with them — the accepted cost. It is then no less reachable than the
tabs it follows: you are already scrolling the strip to see those, and the same
scroll brings the button back. ⌘O, the command palette and the Welcome screen
all still reach the action, none of which the strip owns. Chrome and VS Code
make the same trade.

**2. The separator is gone.** The button's `borderLeft: 1px solid
var(--border-0)` sat against the preceding tab's own `borderRight` and rendered a
double line. The tab's border is now the only divider.

**3. `scrollIntoView` had to change, and this is the interaction decision 1
creates.** The button is the scroller's last child, so scrolling the *active tab*
into view stops at that tab's right edge and leaves the button half-clipped
whenever the last tab is the active one. The effect now aims at the button in
exactly that case: being the final child, revealing it scrolls to the end, which
reveals the tab as well. Measured on the e2e stack with ten tabs and the last one
active — `scrollLeft` 238, button right edge 1200, scroller right edge 1200:
flush, not clipped.

The strip stays chrome throughout: 28px tall, no `var(--row-step)`, not a
`PGPane`, so the button gains no pane-scoped focus behaviour.

## Screenshots

WebKitGTK 605.1.15 under xvfb, 1200x800, captured by a throwaway spec that was
deleted before committing. Ten `pg-e2e-*` temp repositories; only the active one
is ever opened, since the restore is lazy.

**Before — two tabs.** The gap the issue is about.

before, two tabs

**After — one tab.**

after, one tab

**After — two tabs.**

after, two tabs

**After — ten tabs, scrolled to the start.** This is where the decision shows:
the strip overflows (`scrollWidth` 1438 vs `clientWidth` 1200) and the button has
scrolled off with the tabs. Scrolling right, or activating a tab near the end,
brings it back.

after, ten tabs from the start

**After — ten tabs, last one active.** The button is revealed with its tab,
flush against the right edge rather than half-clipped.

after, ten tabs last active

**Before — ten tabs, last one active**, for completeness: at overflow the old
pinned layout and the new one coincide, which is the issue's point restated as a
picture.

before, ten tabs last active

One full-window frame for context (`02-two-tabs-window.png` in the same
directory).

## Tests

`data-testid="repo-tab-new"` is untouched, so **`RepoTabs.test.tsx` passes
unmodified** — no change to that file in this branch.

New coverage:

- `src/design/chrome.test.tsx` — the button is the last tab's
  `nextElementSibling` and shares its parent (inside the scroller, not beside
  it); no `border-left`; and the scroll effect targets the BUTTON when the last
  tab is active, the active TAB otherwise. That last case was confirmed to fail
  against the previous target expression, so it is not vacuous.
- `e2e/specs/repo-tabs.e2e.ts` — the geometry jsdom cannot produce (it performs
  no layout, so every rect there is zero): with two tabs the gap to the last tab
  is ~0 while the distance to the strip's right edge is not (the old layout
  scored the opposite on both), and with the last of ten tabs active the button
  is fully inside the scroller's box. `overflowing` is asserted first as a
  vacuity guard.

Gates, all green: `pnpm tsc --noEmit`, `pnpm test` (193 files, 1653 tests),
`pnpm exec tsc -p e2e/tsconfig.json --noEmit`, and
`pnpm test:e2e:docker run --spec e2e/specs/repo-tabs.e2e.ts` (9 passing) against
a snapshot rebuilt from this branch's code.

## Not verified here

Appearance was checked on WebKitGTK only. The macOS rendering has been confirmed
separately by hand, but nothing in this branch was screenshotted on WKWebView.


jonassaa:
### Measured evidence for decision 3

Read off the live strip on the e2e stack (WebKitGTK 605.1.15 under xvfb,
1200x800), ten repositories open. `plus.parentElement` is the scroller after
this change and the strip itself before it, which is why `overflowing` reads
differently — that difference is itself the placement.

**Before** (button pinned outside the scroller):

```
{"boxRight":1200,"clientWidth":1200,"overflowing":false,"scrollWidth":1200,
 "scrollLeft":0,"plusLeft":1172,"plusRight":1200}
```

Pinned at the window edge regardless of where the tabs end.

**After, ten tabs scrolled to the start** — the strip genuinely overflows and the
button has travelled off with the tabs:

```
{"boxRight":1200,"clientWidth":1200,"overflowing":true,"scrollWidth":1438,
 "scrollLeft":0,"plusLeft":1410,"plusRight":1438}
```

**After, last of ten tabs active** — the effect aimed at the button, so the
scroller went to its end and the button is flush inside the box rather than
half-clipped:

```
{"boxRight":1200,"scrollLeft":238,"plusLeft":1172,"plusRight":1200,
 "fullyVisible":true}
```

Both numbers are now assertions in `e2e/specs/repo-tabs.e2e.ts`
(`gapAfterLastTab`, `plusRightToStripRight`, `plusFullyVisible`), with
`overflowing` asserted first so the second test cannot pass vacuously on a
non-overflowing strip.

### Frames

| state | image |
|---|---|
| before, two tabs |  |
| after, one tab |  |
| after, two tabs |  |
| after, ten tabs from the start |  |
| after, ten tabs last active |  |

PR #219, opened 2026-08-25
feat(repo): name the active repository and branch in the window title

### LINKED ISSUE #217: Window title should name the active repository and branch
The window title is the static string `"PlatypusGit"`
(`src-tauri/tauri.conf.json`, line 15). With several repos open in tabs, or the
app sitting among twenty other windows, the OS window switcher, the dock/taskbar
and Mission Control all show identical entries. Anyone who works across more
than one repository loses the ability to switch by window.

## What to do

Keep the title in sync with the active repository, and ideally the checked-out
branch:

```
myrepo — main — PlatypusGit
```

Put the project name last: window switchers and taskbar tooltips truncate from
the right, so the part that distinguishes windows has to come first.

The pattern already exists in the codebase — the merge resolver sets its own
title in `src/features/merge/MergeWindow.tsx` (~line 154):

```ts
void getCurrentWindow().setTitle(`Resolve: ${path}`);
```

Do the same from wherever the active repo is known. The active tab lives in
`src/features/repo/useTabsStore.ts`, and `useRepoStore` holds exactly one
repository's state (the active tab's) including its current branch. An effect
that reacts to active-repo and branch changes is the shape to aim for.

Cases to handle:

- No repo open (the Welcome screen) — just `PlatypusGit`.
- Detached HEAD — show the short OID rather than an empty branch segment.
- Unborn branch (a fresh `git init` with no commits).
- The title must follow tab switches, branch checkouts, and repo close.

## Tests

`setTitle` is already mocked in `src/test/setup.ts` (~line 44), so a component
test can assert the exact title for each case above without touching Tauri. Add
one; the detached-HEAD and no-repo cases are the ones that will regress.

Good first issue: small, frontend-only, an existing pattern to copy, and the
test seam is already in place.

jadhavgaurav:
## Summary

The window title was the static string `"PlatypusGit"`. With several repos open in tabs, or the app sitting among other windows, the OS window switcher, dock/taskbar, and Mission Control all show identical entries.

- `myrepo — main — PlatypusGit` — on a branch
- `myrepo — a1b2c3d — PlatypusGit` — detached HEAD, short oid instead of an empty segment
- `myrepo — PlatypusGit` — unborn branch (fresh `git init`, no commits yet)
- `PlatypusGit` — no repo open (Welcome)

Repo name first, app name last: window switchers and taskbar tooltips truncate from the right, so the part that distinguishes windows has to lead.

## Why a new backend field

`RepoHandle.head` (set once by `open_repo`) does not follow a checkout within the session — nothing in the app re-fetches it after `checkoutBranch`/`checkoutRef`/rebase etc., only `refreshAll`'s already-live fields (`branches`, `status`, …) do. Relying on it directly for the title would silently go stale on exactly the case the issue calls out as regression-prone (detached HEAD after a checkout).

So this adds `head_info` (`src-tauri/src/commands/repo.rs`), a small read mirroring the branch/oid split `WorktreeInfo`'s own worktree listing already uses (`branch: Option<String>`, `head_oid: Option<String>` — `branch` is `None` on a detached HEAD, `head_oid` is `None` only when unborn). It's called alongside the rest of `refreshAll`'s `Promise.all` in `useRepoStore.ts`, degrading to `null` on failure rather than failing the whole refresh (same policy as `bisect_status`).

Note: `repo.head()`'s `shorthand()` returns the literal string `"HEAD"` on a detached HEAD (not an oid) — confirmed with a throwaway git2 probe — so `is_branch()` is what actually distinguishes "on a branch" from "detached", not just presence of a name.

## Implementation

- `src-tauri/src/git/types.rs` — `HeadInfo { branch, head_oid }`.
- `src-tauri/src/git/mod.rs` / `libgit2.rs` / `cli.rs` — trait method, implementation, `NotImplemented` stub (standard `GitBackend` op pattern).
- `src-tauri/src/commands/repo.rs` + `lib.rs` invoke_handler — `head_info` command.
- `src/lib/types.ts` / `tauri.ts` — TS mirror + wrapper.
- `src/features/repo/repoSlice.ts` / `useRepoStore.ts` — new `headInfo` slice field, refreshed in `refreshAll`.
- `src/features/repo/windowTitle.ts` — pure `windowTitleFor(repoLabel, head)`, so the effect carries no fallback logic of its own.
- `src/AppShell.tsx` — an effect reacting to the active tab's disambiguated label (`useTabsStore`'s existing `labels()` — previously unused anywhere in the app) and `headInfo`, calling `getCurrentWindow().setTitle(...)` the same way `MergeWindow.tsx` already does for its own window.
- `docs/dev/architecture.md` — new command listed under `commands/repo.rs`.

## Tests

- `src-tauri/tests/head_info.rs` — three backend cases: on a branch, detached HEAD, unborn branch.
- `src/features/repo/windowTitle.test.ts` — all five title-format cases as pure-function unit tests.
- `src/AppShell.windowTitle.test.tsx` — component test using the existing `setTitle` mock (`src/test/setup.ts`), covering no-repo, on-branch, detached, unborn, and repo-close.

## Test plan

- `cargo test --manifest-path src-tauri/Cargo.toml --lib` — 157 passed, 0 failed.
- `cargo test --manifest-path src-tauri/Cargo.toml --test head_info` — 3 passed.
- `pnpm tsc --noEmit` — clean.
- `pnpm vitest run` — 215 files / 1853 tests passed (includes doc/tree invariants).

Fixes #217

---
This PR was prepared with AI assistance (Claude); the diff was reviewed and all listed test commands were run and observed to pass before opening.

### DIFF
--- a/docs/dev/architecture.md
+++ b/docs/dev/architecture.md
@@ -99,9 +99,11 @@ git/
                  (user.name/email lookup, NoSignature when unset) and
                  apply_signoff (Signed-off-by trailer, idempotent)
 commands/        Thin Tauri handlers, one file per area:
-├── repo.rs      open_repo, close_repo, trust_repo_path, get_status,
-│                list_all_files, append_gitignore, open_in_editor, and THREE
-│                distinct file readers: read_file_content (working tree),
+├── repo.rs      open_repo, close_repo, trust_repo_path, get_status, head_info
+│                (HEAD's branch/oid, re-polled every refresh — unlike
+│                RepoHandle.head, which open_repo sets once), list_all_files,
+│                append_gitignore, open_in_editor, and THREE distinct file
+│                readers: read_file_content (working tree),
 │                read_file_content_at_rev + list_files_at_rev (a commit's tree),
 │                read_file_content_at_index (the STAGED blob)
 ├── cli.rs       take_launch_intent, cli_shim_status, install_cli_shim
--- a/src-tauri/src/commands/repo.rs
+++ b/src-tauri/src/commands/repo.rs
@@ -4,7 +4,7 @@ use tauri::State;
 
 use crate::{
     error::{AppError, AppResult},
-    git::types::{FileContent, FileStatus, RepoHandle, RepoId},
+    git::types::{FileContent, FileStatus, HeadInfo, RepoHandle, RepoId},
     state::AppState,
 };
 
@@ -41,6 +41,18 @@ pub async fn trust_repo_path(state: State<'_, AppState>, path: String) -> AppRes
     let backend = state.backend.clone();
     let path_buf = PathBuf::from(path);
     tokio::task::spawn_blocking(move || backend.trust_path(&path_buf))
+        .await
+        .map_err(|e| AppError::Internal(e.to_string()))?
+}
+
+/// HEAD's current branch/oid, meant to be re-polled on every refresh — unlike
+/// `RepoHandle.head`, which `open_repo` sets once and the frontend must not
+/// treat as live after a checkout (#217).
+#[tauri::command]
+pub async fn head_info(state: State<'_, AppState>, repo_id: String) -> AppResult<HeadInfo> {
+    let backend = state.backend.clone();
+    let repo_id = RepoId(repo_id);
+    tokio::task::spawn_blocking(move || backend.head_info(&repo_id))
         .await
         .map_err(|e| AppError::Internal(e.to_string()))?
 }
--- a/src-tauri/src/git/cli.rs
+++ b/src-tauri/src/git/cli.rs
@@ -6,7 +6,7 @@ use super::{
     types::{
         AheadBehind, BisectMark, BisectStatus, BlameLine, BranchInfo, CommitInfo, CommitOptions, ConflictSides,
         DiffKind, FileContent,
-        FileDiff, FileStatus, LfsStatus, LogFilter, LogPage, RebaseStatus, RebaseStep, ReflogEntry,
+        FileDiff, FileStatus, HeadInfo, LfsStatus, LogFilter, LogPage, RebaseStatus, RebaseStep, ReflogEntry,
         RemoteInfo, RepoHandle, RepoId, RepoState, ResetMode, StashInfo, StashSaveOptions,
         SubmoduleInfo, TagInfo,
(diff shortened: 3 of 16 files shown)


jonassaa:
Thanks for this — the `head_info` addition is better justified than the issue's "frontend-only" framing suggested. I checked whether the title could be driven from existing live state, and it can't: `currentBranch(branches)` is `null` on a detached HEAD, and the paged log isn't a valid source for HEAD questions, so there genuinely is no live short-oid in the frontend today. Adding the backend read was the right call.

One change I'd like before merge, in `src/AppShell.tsx`:

```ts
const tabs = useTabsStore((s) => s.tabs);
```

This subscribes the top-level `AppShell` to the entire tab array, so a write to *any* tab now re-renders the whole shell — titlebar, `RepoTabs`, the dialogs, `AppBody`, `CommandPalette`. `refreshBadges()` runs on every window focus and calls `patchTab` once per inactive open tab, so with four tabs open, alt-tabbing back into the app triggers four full-tree re-renders where it previously only re-rendered the tab strip.

This file already records the same lesson a few hundred lines down (~676), about a selector-less `useRepoStore()` that "re-rendered the whole titlebar (BranchChip, UpdateChip, picker) on EVERY store write".

Deriving the label inside the selector keeps the subscription to a *string*, which zustand compares by value — so it re-renders only when the label actually changes, and `activePath` drops out of the deps since the selector reads it:

```ts
const repoLabel = useTabsStore((s) => {
  const i = indexOfTab(s.tabs, s.activePath);
  return i < 0 ? null : labelTabs(s.tabs)[i] ?? null;
});
const headInfo = useRepoStore((s) => s.headInfo);
React.useEffect(() => {
  void getCurrentWindow().setTitle(windowTitleFor(repoLabel, headInfo));
}, [repoLabel, headInfo]);
```

`indexOfTab` and `labelTabs` are both already exported from `src/features/repo/tabs.ts` (`RepoTabs.tsx:27` uses `labelTabs` the same way). Using `indexOfTab` also drops the hand-rolled `t.path === activePath` comparison in favour of the normalising helper.

I applied this locally against your branch: `pnpm tsc --noEmit` clean and `pnpm vitest run` green at 215 files / 1853 tests, including your five new window-title cases unchanged.

PR #222, opened 2026-08-25
feat(shell): name the active repository and branch in the window title

### LINKED ISSUE #217: Window title should name the active repository and branch

SurajPatelPro:
## Summary
- The window title was always the static `"PlatypusGit"`, so window switchers, the dock/taskbar and Mission Control showed identical entries for every open repo/window.
- Added `windowTitle` (`src/lib/derive.ts`), a pure function composing `name — branch — PlatypusGit` (project name last, so it survives the right-truncation window switchers apply):
  - On a branch: the branch name.
  - Detached HEAD: falls back to the short OID, via the same `RepoHandle.head` field `BranchChip` already reads the same way (branch shorthand when on a branch, short OID when detached, `null` when unborn).
  - Unborn branch (fresh `git init`, no commits): branch segment omitted rather than shown empty.
  - No repo open: just `PlatypusGit`.
- `AppShell` derives the title from `useRepoStore`'s `current`/`branches` and syncs it via `getCurrentWindow().setTitle` in an effect — no new store plumbing, so it follows tab switches, branch checkouts and repo close automatically (same mechanism `MergeWindow.tsx` already uses for its own title).

Closes #217

## Test plan
- [x] `pnpm tsc --noEmit` — clean
- [x] Added `src/lib/derive.test.ts` — unit coverage for all four `windowTitle` cases (named branch, detached/short-OID, unborn/no-commits, no-repo)
- [x] Added `src/AppShell.windowtitle.test.tsx` — component coverage asserting `getCurrentWindow().setTitle` is called correctly on mount, on opening a repo, for detached HEAD, for an unborn branch, and on tab close (mirrors the existing `setTitle` mocking `MergeWindow.test.tsx` already uses)
- [x] Full suite: `pnpm test` — 215 test files, 1853 tests, all passing (no regressions)
- [ ] Did not run `pnpm test:e2e:docker` (Docker-based e2e) — out of scope per the issue ("small, frontend-only... a component test can assert the exact title... without touching Tauri") and not available in my sandbox; flagging honestly rather than claiming it.

### DIFF
--- a/src/AppShell.tsx
+++ b/src/AppShell.tsx
@@ -1,4 +1,5 @@
 import { listen } from "@tauri-apps/api/event";
+import { getCurrentWindow } from "@tauri-apps/api/window";
 import React from "react";
 import {
   PGActivityBar,
@@ -74,6 +75,7 @@ import {
   isStaged,
   isUnstaged,
   totalAheadBehind,
+  windowTitle,
 } from "@/lib/derive";
 
 export type ScreenId =
@@ -162,8 +164,19 @@ export function AppShell() {
   usePreventBrowserContextMenu();
   useCliLaunch();
   const repo = useRepoStore((s) => s.current);
+  const branches = useRepoStore((s) => s.branches);
   const error = useRepoStore((s) => s.error);
   const clearError = useRepoStore((s) => s.clearError);
+
+  // Keep the OS window title naming the active repository and branch (#217):
+  // window switchers, the dock/taskbar and Mission Control otherwise show
+  // identical "PlatypusGit" entries for every open repo. `windowTitle` already
+  // covers detached HEAD (short OID) and unborn (repo name only); this effect
+  // just follows whatever it derives across tab switches, checkouts and close.
+  const title = windowTitle(repo?.path ?? null, branches, repo?.head ?? null);
+  React.useEffect(() => {
+    void getCurrentWindow().setTitle(title);
+  }, [title]);
 
   // Launch always lands on History — it is the screen that answers "what is
   // going on in this repo", so it is worth more than restoring wherever the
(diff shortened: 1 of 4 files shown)


jonassaa:
#217 already closed by #219

Issue #35, opened 2026-07-03
e2e: Linux CI roaming flake — hung evals hit 30s script timeout, retry double-runs side-effectful scripts


jonassaa:
## Symptom

The `e2e-linux` job intermittently fails on a *random* spec — a different test each time — always with the same signature in the log:

```
WARN webdriver: WebDriverError: Script execution timed out when running "execute/sync" with method "POST"
...
Error: Can't call elementClick on element with selector "..." because element wasn't found
```

Observed occurrences (all 2026-07-03):

- `main` push run at 10:35 (post-#33 squash): `history danger ops.reset soft moves HEAD and keeps changes staged` — **pre-dates #34**, so not introduced by it.
- PR #34 runs 1 & 2: `merge & conflict.resolves with accept-ours and finalizes the merge` (twice).
- PR #34 run 3 (plain rerun, no code change): green. Confirms roaming flake, not deterministic failure.

## Mechanism (established while debugging #34)

`tauri-plugin-wdio-webdriver` executes scripts via `evaluateJavaScript` and waits `tokio::time::timeout(script_ms)` (default 30s) for the completion handler (`src/platform/linux.rs`, same shape as macOS). Two failure modes:

1. **Dropped handler** (macOS-typical): eval fired mid-`refresh()` navigation never completes → driver burns the full script timeout → error → caller retries. Pure latency; retry is safe because the script never ran.
2. **Late completion** (Linux-CI-typical): under xvfb the eval *does* run but slower than the timeout → driver reports timeout → WebdriverIO retries → **the script executes twice**. For side-effectful helpers (`jsContextMenu`, `jsClickMenuItem`, dialog stubs, palette key dispatch) the double-run desyncs the UI flow, and a later element (e.g. `[data-testid="mark-resolved"]`) legitimately never appears.

This is why #34's fix is platform-gated: `browser.setTimeout({ script: 2500 })` converts mode 1 into an invisible 2.5s retry on macOS, but on Linux it would *increase* mode 2's double-run frequency. Linux keeps the 30s default and keeps the flake.

## Fix directions (not mutually exclusive)

- **Idempotent helpers:** make every side-effectful `browser.execute` script safe to re-run (guard flags on `window`, e.g. skip if the context menu is already open / the stub is already installed / the keydown was already dispatched for this token). Cheap, local, addresses mode 2 regardless of driver behavior.
[...]

PR #223, opened 2026-08-25
fix(e2e): settle the reload race that owned 70-80% of e2e wall time

### LINKED ISSUE #194: 70-80% of e2e wall time is 30s script-timeout stalls, not test work
Measured while sharding the suite (issue 189). The e2e suite's cost is dominated by a single repeating failure mode, and it is not the specs.

## The measurement

Stall count tracks total spec time almost linearly, at ~30s each, across three consecutive `main` runs:

| run | specs | spec time | `Script execution timed out` |
|---|---|---|---|
| `fb630a5` | 26 | 9m19s | 13 |
| `65ecf81` | 27 | 11m56s | 18 |
| `78492a4` | 26 | 13m55s | 23 |

13 × 30s = 6m30s of the 9m19s. 23 × 30s = 11m30s of the 13m55s. So **roughly 70–80% of e2e wall time is the driver waiting out a script timeout**, and which specs pay moves run to run because it is the reload race, not the spec.

The per-spec signature is unmistakable once you look for it: every spec over 30s comes out as `n × 30` plus small change (33.2s, 32.8s, 34.1s, 32.6s, 1m36s, 2m07s) while specs with few refreshes finish in milliseconds to single-digit seconds.

## Why the obvious fix is already refused

`e2e/wdio.conf.ts` caps the W3C script timeout at 2500ms **on macOS only**, and the comment says why not on Linux:

> Do NOT apply on Linux CI: under xvfb, legitimate executes can exceed 2.5s, and a timed-out-but-completed script gets retried — double-running side-effectful helpers (observed: merge-conflict flow desync).

So a blanket cap trades slowness for a correctness bug, which is a worse trade. Any fix has to deal with the retry-double-run hazard, not just the timeout.

## Directions worth exploring

- **A middle cap.** 2.5s is too tight for xvfb; 30s is the whole problem. Something like 8–10s bounds the stall while leaving legitimate executes room. Needs measuring what the *slowest legitimate* execute under xvfb actually is, and it does not by itself fix the double-run risk.
- **Make the retried helpers idempotent**, which removes the hazard the cap is refused for and makes any cap safe. Probably the highest-value direction: it attacks the reason rather than the symptom.
- **Attack the race itself.** The stall is an `execute()` fired while a `refresh()` navigation is mid-document-swap, whose completion handler is dropped. Fewer refreshes means fewer rolls — issue 189 already measured that *adding* one refresh per spec file took the suite from 528s to 1064s. `openRepo()` calls a refresh per `it()` (`keymap` 28 times); a settle-gate that provably waits for the new document before any `execute()` would remove the roll rather than shortening it.
- **Count the stalls in CI** and fail or warn above a threshold, so a regression here is visible instead of showing up as "e2e is slow again".

## Why it matters more than sharding

Sharding parallelises across runners but cannot make a stall shorter — it only redistributes the dice. At 13 stalls the floor is ~3 minutes of real work; the other 6+ minutes are recoverable without touching a single spec. This is the cheapest large win available on CI time.

jonassaa:
Closes #194.

## What was actually costing the time

An `execute()` that lands while a `browser.refresh()` navigation is
mid-document-swap loses its completion handler. The driver then waits out the
**entire** W3C script timeout before erroring — 30 s, uncapped on Linux CI — so
a loss costs the *timeout*, not the work.

Every refresh site in the tree fired `armDriverBridge()` — an `execute()` — as
its **first** post-refresh command. That is one roll of that die per refresh,
and `openRepo()` refreshes once per `it()` (`keymap`: 28×). Which is why #194
found stall count tracking refresh count rather than spec complexity, and why
adding one refresh per spec file once took the suite from 528 s to 1064 s.

## The fix, in the order it does the work

1. **`refreshAndSettle` owns every `browser.refresh()`** (`e2e/support/app.ts`):
   refresh → a matched WebDriver find → *then* arm. A find issued mid-swap
   either matches (the driver's own proof navigation settled) or misses and is
   re-polled for pennies. The pre-find arm it replaces could only ever land on
   the *dying* document — which is exactly why every call site already armed a
   second time afterwards. Deleting it costs nothing and removes the roll.
2. **The Linux script timeout is capped (8 s; macOS keeps 2.5 s).** The old
   refusal was correct when written — a timed-out-but-completed script gets
   retried and double-runs side-effectful helpers (the merge-conflict desync,
   #35) — but `executeOnce` closed that hole afterwards, makes a retry a no-op
   returning the first run's value, and is self-tested in `harness.e2e.ts`. I
   re-audited every `browser.execute` in `e2e/`: the side-effectful ones all go
   through `executeOnce`; the remaining bare ones (localStorage seeds,
   `scrollTop`, a `Selection`, `window.close`) are idempotent by construction.
   8 s is measured — the suite's scripts run **p50 6 ms / p99 12 ms / max 12 ms**
   under xvfb.
3. **Stalls are counted and printed per spec file** (`e2e/support/scriptTiming.ts`),
   so a regression is attributable instead of "e2e is slow again".

## Measured, in Docker — same WebKitGTK+xvfb stack as CI, same machine

| | wall | stalls | result |
|---|---|---|---|
| whole suite, before | 6 m 18 s | 9 | 28/28 |
| whole suite, after | **1 m 55 s** | **0** | 28/28 |
| `keymap`+`repo-tabs`, before | 1 m 15 s | 2 | green |
| `keymap`+`repo-tabs`, after | **16 s** | **0** | green |

9 × 30 s is 71 % of the 6 m 18 s baseline, so the arithmetic closes on what the
issue measured. The `keymap`+`repo-tabs` pair was run with the script timeout
**held at 30 s** so the gate is attributed on its own rather than to the cap —
the cap is only the belt for a refresh site that ever slips past the rule.

## What keeps it fixed

A rule kept by comments is what decayed the first time, so both halves are
static tests at `pnpm test` speed:

- **`test/e2eRefreshGate.test.ts`** — `browser.refresh()` may be called from
  exactly one place, and every gate handed to it must **start** with a WebDriver
  query. The second half matters on its own: a gate that polls in page obeys the
  first rule and reinstates the stall anyway. Mutation-checked — both halves
  fail on a planted violation.
- **`test/scriptTiming.test.ts`** — pins the cap resolver. The obvious spelling
  (`Number(env.X ?? default)`) reads compose's empty-string passthrough as a
  **zero** timeout, under which the driver fails every `element`, `elements` and
  click instantly. That is not hypothetical: it bit this branch mid-development
  and read as a wholly broken session.

## Verification

- `pnpm test` — 217 files, 1865 tests, green
- `pnpm tsc --noEmit` and `pnpm exec tsc -p e2e/tsconfig.json --noEmit` — clean
- `pnpm test:e2e:docker` full suite — 28/28, 1 m 55 s, 0 stalls

Docs updated where the rules live: `docs/dev/testing.md` and the `e2e-testing`
skill.


jonassaa:
CI on this branch (run 32877754926) — all green, **0 stalls across all 28 specs**:

| shard | specs | spec time |
|---|---|---|
| 1 | 9 | 47s |
| 2 | 3 | 58s |
| 3 | 9 | 47s |
| 4 | 7 | 32s |

**3m04s** of total spec time, against the three `main` runs #194 measured: 9m19s (26 specs), 11m56s (27), 13m55s (26). Slowest single in-page script anywhere on the runner: 190ms — so the 8s cap has ~40x headroom, and the per-spec timing line reports it without anyone having to go looking.

The stall count that used to be 13–23 per run is 0.

PR #231, opened 2026-08-26
Add reveal in file manager and open in terminal to context menus

### LINKED ISSUE #215: Reveal in Finder / Explorer and Open in terminal from file and repo menus
You can copy a file's path to the clipboard from most surfaces in the app
(`src/design/context-menu.tsx` does this in a dozen places), but you cannot ask
the OS to show it. "Reveal in Finder" and "Open in terminal here" are in every
competing client and are the two commands people reach for when they want to
leave the GUI for a moment.

## What to add

Two context-menu actions, on the file rows in the Commit panel and repo browser
(next to the existing "Copy path"), and on the repository itself — the repo tab
right-click menu (`src/features/repo/RepoTabs.tsx`, which already offers "copy
path") is the natural home for the repo-level version.

- **Reveal in file manager** — open the OS file manager with the file selected.
- **Open in terminal** — open a terminal at the containing directory.

Label them per platform: "Reveal in Finder" (macOS), "Show in Explorer"
(Windows), "Show in file manager" (Linux). `src/lib/platform.ts` already exists
for this.

## Implementation notes — read these first

- **`Command::new` is banned outside `src-tauri/src/proc.rs`** and a guard test
  fails the build if you try. Use the `proc::program*` constructors. See
  `docs/dev/backend.md`.
- New commands go in `src-tauri/src/commands/repo.rs`, stay thin, and must be
  registered in `invoke_handler![…]` in `src-tauri/src/lib.rs`.
- **You must add the new commands to the `commands/repo.rs` entry in the backend
  tree in `docs/dev/architecture.md`** — `test/docs.test.ts` fails the build
  otherwise.
- Add the TS types to `src/lib/types.ts` and typed wrappers to `src/lib/tauri.ts`
  (the frontend never calls `invoke()` directly).
- Check `src-tauri/capabilities/default.json` for whatever permission this needs.

Platform commands:

| | Reveal | Terminal |
|---|---|---|
| macOS | `open -R <path>` | `open -a Terminal <dir>` |
| Windows | `explorer /select,<path>` | `wt -d <dir>`, falling back to `cmd /c start` |
| Linux | `xdg-open <dir>` (selection is not portable) | `x-terminal-emulator` / `$TERMINAL`, with fallbacks |

Traps worth knowing before you start:

- **`explorer.exe` returns exit code 1 on success.** Do not treat non-zero as
  failure on Windows.
- **A console window must not flash on Windows** — this bit us before (#172).
  Follow whatever `proc.rs` already does for `CREATE_NO_WINDOW`.
- **Linux terminal detection is genuinely unreliable.** Try a short ordered list
  and fail with a clear error rather than silently doing nothing. Making the
  terminal command a setting is a reasonable follow-up, not part of this.
- Pass user paths safely — end option parsing with `--` where the tool supports
  it, and never build a shell string.

## Tests

Rust unit tests over the argv builders per platform (pure functions, no spawning
— that is the pattern the repo already uses for command construction), plus a
component test that the menu items appear and call the right wrapper.

Good first issue in scope, but it touches the full stack — a nice way to learn
the codebase end to end. Read `docs/dev/backend.md` and the "Adding a new git
op" recipe in `CONTRIBUTING.md` first.

jadhavgaurav:
## Summary

Adds the two context-menu actions the issue asks for: **Reveal in
Finder/Explorer/file manager** and **Open in terminal**, on file rows in the
Commit panel and repo browser (next to the existing "Copy path"), and on the
repo tab strip's menu (the repo-level version).

## Implementation

- `src-tauri/src/reveal.rs` (new): builds the per-platform argv as **pure**
  functions decoupled from the actual host via a `HostPlatform` enum —
  `HostPlatform::current()` is the only place `cfg!(target_os = …)` is read,
  mirroring `opener.rs::opener_program`'s own runtime-`cfg!()` pattern — so
  `reveal_plan`/`terminal_plan` are unit-tested for all three platforms from
  any host, with no spawning in the tests.
  - macOS: `open -R <path>` / `open -a Terminal <dir>`.
  - Windows: `explorer.exe /select,<path>` (or the plain dir when there's no
    file to select) / `wt.exe -d <dir>`, falling back to `cmd.exe` when
    Windows Terminal isn't installed.
  - Linux: `xdg-open <parent-dir>` (no portable "select this file" verb) /
    an ordered terminal-candidate list (`$TERMINAL`, then
    `x-terminal-emulator`, `gnome-terminal`, `konsole`, `xfce4-terminal`,
    `xterm`), failing with a clear error naming what was tried rather than
    doing nothing silently.
  - `explorer.exe`'s exit code 1 **on success** is handled explicitly, not
    treated as failure.
  - Every spawn goes through `proc::program_async` (issue 172) — no raw
    `Command::new`, so the `CREATE_NO_WINDOW`/no-console-flash guarantee
    holds here too. The terminal launch uses `.spawn()`, not
    `.status().await`, since a terminal is long-running and awaiting its
    exit would block the app.
  - No shell interpreter anywhere — argv only, `--`-safe where relevant.
- `commands/repo.rs`: two thin commands, `reveal_in_file_manager` and
  `open_in_terminal`, both `repo_id` + optional `relative_path` — reusing
  `open_in_editor`'s exact shape and `opener::safe_workdir_path` for the
  escape check. An omitted/empty `relative_path` targets the repo's root
  directory instead of a file, which is what the repo tab's menu passes.
  Registered in `lib.rs`'s `invoke_handler!`.
- `src/lib/types.ts`/`tauri.ts`: no new `AppError` variants needed (spawn
  failures reuse `Io`); typed wrappers `revealInFileManager`/`openInTerminal`
  added to `tauri.ts`, the frontend's only `invoke()` boundary.
- `useRepoStore.ts`: two new actions mirroring `openInEditor`'s shape, for the
  file-row menus (always the active repo).
- `context-menu.tsx`: `fileMenuItems` grows an optional `platform` parameter
  and the two new items, right after "Copy path". `RepoTabs.tsx`'s
  `tabMenuItems` gets the repo-level pair, gated on the right-clicked tab
  actually having a `repoId` (a still-loading/failed tab has nothing to
  resolve a directory from, so the entries are disabled rather than firing at
  a repo that isn't open).
- `src/lib/platform.ts`: `fileManagerLabel(platform)` for the per-platform
  wording ("Reveal in Finder" / "Show in Explorer" / "Show in file manager"),
  reused by both menu builders so the wording can't drift between them.
- `docs/dev/architecture.md`: documented `reveal.rs` and the two new
  `commands/repo.rs` entries (`test/docs.test.ts` enforces this).

## Scope

- Embedded-repo rows (`embeddedRepoMenuItems`) and submodule rows are **not**
  touched — the issue names file rows and the repo tab specifically, and
  those two builders are already distinct menus with their own scope.
- **The terminal command is not user-configurable.** The issue calls this out
  explicitly as "a reasonable follow-up, not part of this [issue]", so it's
  left as a fixed fallback list on Linux and the obvious default elsewhere.
- No new Tauri capability was needed in `capabilities/default.json` — these
  are ordinary custom `#[tauri::command]`s (not Tauri plugin commands), same
  as every other command in `commands/repo.rs`.

## Tests

- Rust: `reveal.rs`'s `#[cfg(test)]` module — 11 pure unit tests over
  `reveal_plan`/`terminal_plan` for all three platforms (selection vs.
  directory targets, the Windows fallback chain, the Linux `$TERMINAL`
  override and its empty-string edge case), no spawning.
- Frontend: `src/design/context-menu.reveal.test.tsx` — the two menu items
  appear with the right per-platform label, dispatch to the right
  `lib/tauri.ts` wrapper with the right `repoId`/`relativePath`, no-op when
  there's no path, and the repo-tab entries disable for a tab with no
  `repoId` yet. Plus `fileManagerLabel` unit tests in `platform.test.ts`.

## Test plan

- `cargo check --manifest-path src-tauri/Cargo.toml` — clean.
- `cargo test --manifest-path src-tauri/Cargo.toml` — 182 passed, 0 failed
  (11 new, in `reveal::tests`), including the `Command::new` guard test.
- `pnpm tsc --noEmit` — clean.
- `pnpm vite build` — clean.
- `pnpm vitest run` — 219 files / 1888 tests passed (includes the
  doc/tree invariants).
- Manual click-through wasn't possible in this (headless/sandboxed) build
  environment; reasoned through each platform branch instead, and the
  Windows/Linux paths are exercised only by the pure argv-builder tests
  above (no CI runner for those OSes locally). macOS's `open -R`/`open -a
  Terminal` path is the one most directly checkable by inspection against
  documented `open(1)` behavior.

Fixes #215

---
This PR was prepared with AI assistance (Claude); the diff was reviewed and all listed test commands were run and observed to pass before opening.

### DIFF
--- a/docs/dev/architecture.md
+++ b/docs/dev/architecture.md
@@ -24,9 +24,11 @@ reveal.rs        "Reveal in Finder/Explorer" + "Open in terminal" (#215).
                  HostPlatform decouples argv building from the actual host, so
                  reveal_plan/terminal_plan are PURE and unit-tested for all
                  three platforms from any host. Spawns via proc.rs only.
-                 explorer.exe's exit-1-on-success is handled here, not treated
-                 as failure; a missing Linux terminal falls through an ordered
-                 candidate list rather than failing silently
+                 explorer.exe's exit-1-on-success is carried as a FLAG on the
+                 plan (not a program-name compare — the program is an absolute
+                 pinned path); Windows launchers are pinned to System32 /
+                 WindowsApps against binary planting; a missing Linux terminal
+                 falls through an ordered candidate list rather than silently
 forge/           GitHub/GitLab PR/MR integration (#92). Trait = URL builders +
                  response parsers, pure and testable against recorded JSON:
 ├── mod.rs       Types + Forge trait, forge_for(kind), injection guards
--- a/docs/dev/architecture.md
+++ b/docs/dev/architecture.md
@@ -20,6 +20,13 @@ proc.rs          THE only sanctioned child-process spawner (issue 172):
                  git / git_async / git_async_in (CREATE_NO_WINDOW,
                  GIT_TERMINAL_PROMPT=0, stdin closed), program / program_async,
                  and the two *_keeping_console exceptions. See backend.md
+reveal.rs        "Reveal in Finder/Explorer" + "Open in terminal" (#215).
+                 HostPlatform decouples argv building from the actual host, so
+                 reveal_plan/terminal_plan are PURE and unit-tested for all
+                 three platforms from any host. Spawns via proc.rs only.
+                 explorer.exe's exit-1-on-success is handled here, not treated
+                 as failure; a missing Linux terminal falls through an ordered
+                 candidate list rather than failing silently
 forge/           GitHub/GitLab PR/MR integration (#92). Trait = URL builders +
                  response parsers, pure and testable against recorded JSON:
 ├── mod.rs       Types + Forge trait, forge_for(kind), injection guards
@@ -102,8 +109,11 @@ commands/        Thin Tauri handlers, one file per area:
 ├── repo.rs      open_repo, close_repo, trust_repo_path, get_status, head_info
 │                (HEAD's branch/oid, re-polled every refresh — unlike
 │                RepoHandle.head, which open_repo sets once), list_all_files,
-│                append_gitignore, open_in_editor, and THREE distinct file
-│                readers: read_file_content (working tree),
+│                append_gitignore, open_in_editor, reveal_in_file_manager,
+│                open_in_terminal (#215 — both take an optional relative_path;
+│                omitted/empty targets the repo ROOT instead, for the repo
+│                tab's menu), and THREE distinct file readers:
+│                read_file_content (working tree),
 │                read_file_content_at_rev + list_files_at_rev (a commit's tree),
 │                read_file_content_at_index (the STAGED blob)
 ├── cli.rs       take_launch_intent, cli_shim_status, install_cli_shim
--- a/docs/dev/backend.md
+++ b/docs/dev/backend.md
@@ -276,6 +276,15 @@ Part of the `docs/dev/` set (`architecture`, `testing`, `frontend`, `backend`,
   process holding the file forever. A stray console window is cosmetic; an
   invisible editor is Task Manager. The guard test allow-lists both call sites
   by name.
+- **Pin Windows system executables to an absolute path.** `CreateProcess`
+  searches the **current directory before the system directory**, and this
+  app's cwd *is* a repository whenever the `pgit` shim launched it from inside
+  one — so a bare `rundll32.exe` / `explorer.exe` / `cmd.exe` lets a cloned
+  repo ship its own copy and have it run. `opener.rs::opener_program` pins via
+  `%SystemRoot%\System32`; `reveal.rs::system32_exe` does the same, and
+  `wt.exe` (an App Execution Alias, not a System32 binary) is pinned under
+  `%LOCALAPPDATA%\Microsoft\WindowsApps` or skipped. Unresolvable beats
+  relative: a missing pin falls back to the next candidate.
 - **Decide with libgit2 before you spawn:** `verify_commit` reads the `gpgsig`
   header via `extract_signature` and answers `SigState::None` with no
   subprocess (same pre-check `verify_tag` has) — a cost win on every platform,
--- a/src-tauri/src/commands/repo.rs
+++ b/src-tauri/src/commands/repo.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use tauri::State;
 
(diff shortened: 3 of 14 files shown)


jonassaa (commented):
Reviewed this properly — it's careful work, and the design choice that stands out is making `reveal_plan`/`terminal_plan` pure functions of a `HostPlatform` rather than `#[cfg]`-gated bodies. That's what lets all three platforms' argv be tested from a macOS dev box, and it's the same reasoning `opener.rs::opener_program` already used. `proc::` used throughout, `safe_workdir_path` reused for the escape check, `explorer.exe`'s exit-1-on-success accounted for, docs updated so `test/docs.test.ts` stays satisfied. CI is green.

Three things need fixing before this lands. All three are invisible to the test suite, which is why I want to call them out rather than just patch them quietly.

### 1. Bare `explorer.exe` / `cmd.exe` / `wt.exe` reopens a binary-planting hole (the blocker)

`opener.rs::opener_program` pins `rundll32.exe` to `%SystemRoot%\System32`, and the comment above it explains why:

> `CreateProcess` searches the **current directory before the system directory**, so a bare `rundll32.exe` lookup is a binary-planting hole — a cloned repository containing its own `rundll32.exe` would be run instead of the real one, and the app's cwd *is* the repository whenever it was launched by the `pgit` shim from inside one.

`reveal.rs` spawns bare `explorer.exe` and bare `cmd.exe`, which is exactly the lookup that pin exists to remove. `open_terminal` widens it further by setting the repository as the child's `current_dir` — so for "Open in terminal" the repo isn't merely *sometimes* the cwd, it always is. A cloned repo shipping its own `cmd.exe` at the root would be what runs.

This one is genuinely hard to see, because the rule lived only in a comment in a different file — nothing in `docs/dev/` recorded it. That's a gap on our side, not yours, and I've now written it into `docs/dev/backend.md` so the next person adding a spawn site finds it.

`wt.exe` is the interesting case: it isn't a System32 binary at all, it's an App Execution Alias the Store install drops in `%LOCALAPPDATA%\Microsoft\WindowsApps`. So it gets pinned there, and when `%LOCALAPPDATA%` is unresolvable the candidate is **skipped** rather than looked up relatively — falling through to the `cmd.exe` fallback you already built. Unresolvable beats relative: a missing pin costs a nicer terminal window, a planted one costs the machine. The trade-off is that a non-Store Windows Terminal install (scoop, portable) now falls back to `cmd.exe`; that seemed clearly the right side to err on.

### 2. `shown != "explorer.exe"` stops matching the moment the program is a path

```rust
if !status.success() && shown != "explorer.exe" {
```

This is correct today and breaks silently the instant fix #1 lands, because `shown` becomes `C:\Windows\System32\explorer.exe`. Every *successful* reveal on Windows would then be reported to the user as an error. Worth decoupling on its own merits regardless — it's now an `exit_status_meaningless` flag on `SpawnPlan`, so the fact travels with the plan that owns it instead of being re-derived from a string.

### 3. macOS ignores `is_dir`, contradicting `reveal_plan`'s own doc

The doc comment says `is_dir` distinguishes "reveal a FILE (select it inside its parent's window)" from "reveal a DIRECTORY (open a window on it directly — the repo-tab menu's case)". Windows and Linux both honour that; macOS emits `open -R` either way, and `open -R <dir>` *selects* the directory in its parent window rather than opening it. So the repo tab's "Reveal in Finder" did something different from "Show in Explorer" for the same click. Now `open -R` for files, plain `open` for directories.

---

### Non-blocking, your call

- **macOS hardcodes `Terminal.app` and ignores `$TERMINAL`, which the Linux branch honours.** Configurability is explicitly out of scope per #215 and I'm not asking for it here — but honouring `$TERMINAL` on macOS too would cost one line and isn't really "configuration", it's just not ignoring a variable you already read.
- **The `cmd /C start` fallback does put a shell interpreter in the chain,** which the module doc's "No shell interpreter is ever involved" claim overstates. It isn't injectable — every argv element is a literal and the directory arrives via `current_dir`, not argv — so I only softened the comment to say so explicitly, rather than restructuring it. `start` is genuinely what gives the new `cmd.exe` a console of its own.
- **Embedded-repo rows carry a trailing-slash path** (`st?.path`, per the comment in `RepoBrowser.tsx`), so on Linux `Path::parent()` of `/repo/sub/` is `/repo` — reveal and terminal land one level up for that row type only. Small enough to leave.
- **No e2e coverage**, which is the right call — you can't spawn a file manager in the CI container. The pure-planner tests are the testable surface and you covered it.

### Verification

`cargo test` exits 0 across 63 test binaries (`reveal`'s unit tests 11 → 14, including one that fails if the System32 pin is removed), `pnpm vitest run` is 219 files / 1888 tests green including the doc invariants, `tsc --noEmit` clean, and the `Command::new` guard test still passes.

The fix is on `fix/reveal-windows-launcher-pinning` upstream — one commit that fast-forwards this branch's tip, so it can go straight on top without a rebase. Happy to push it here if you'd rather I did; otherwise cherry-picking it keeps this as one clean squash merge under your name.

Thanks for this — the `HostPlatform` split in particular is the shape I'd want the next platform-branching module to copy.

jonassaa:
Merged. For the record: the three fixes from the review above went in as d548f69 on this branch just before the merge, so the squash on `main` carries both commits and their full messages. I pushed it here rather than as a separate follow-up PR so this stays one clean squash — but that does mean the squash is authored under your name while containing my changes, hence this note rather than a silent commit appearing.

Two of the non-blocking items are still open if you ever want a follow-up:

- honouring `$TERMINAL` on macOS the way the Linux branch already does, and
- the embedded-repo trailing-slash path (`st?.path`) landing one directory up on Linux, since `Path::parent()` of `/repo/sub/` is `/repo`.

Neither is urgent and neither blocks anything.

Thanks for this — the `HostPlatform` split is the part I'd point at as the model for the next platform-branching module we add. Closes #215.

PR #262, opened 2026-08-26
Superseded by #260 — cancel-net-ops (op-id design + extras, kept for reference)

### LINKED ISSUE #234: Cancel a running clone, fetch, pull or push
A clone, fetch, pull or push that hangs can only be escaped by force-quitting the
app. There is no cancel button and no timeout.

Our own code already admits it — `src-tauri/src/commands/create.rs:253`:

> *"— and a clone has no cancel button, so a hang here is force-quit"*

and again at `:259` on what happens when the window closes mid-clone.

## Why

Cloning `torvalds/linux` over a slow link, or hitting a host that accepts the TCP
connection and then stalls, is not an exotic case — it is a first-five-minutes
case, and force-quit is the worst possible first impression.
[GitHub Desktop #2082](https://github.com/desktop/desktop/issues/2082) ("Canceling
the cloning of a repo?") has been open for years.

#212 lists "does progress report, can it be cancelled, does the UI stay
responsive?" in its clone checklist, and #212's own rule is that every failure
found gets its own issue. This is that issue.

## Where

- `src-tauri/src/commands/create.rs` — `run_clone`, `clone_repo`
- `src-tauri/src/commands/net.rs` — `run_git_authenticated` (fetch/pull/push),
  including the auto-fetch timer, which can stack up stalled fetches
- `src/features/create/CloneDialog.tsx` — where the cancel affordance goes

## How

Network git ops are spawned child processes (`proc.rs`), so cancellation is a kill
of the child plus cleanup:

- Keep a handle to the in-flight child, keyed so the frontend can cancel a
  specific operation.
- Cancel = kill the process group, then clean up. **A cancelled clone must not
  leave a half-written target directory behind** — decide explicitly whether to
  remove a directory we created (and never one that already existed; see
  `validate_clone_target`).
- Cancel on window close too, per the `:259` comment.
- Give the long ops a visible cancel button and real progress. A cancelled op must
  surface as cancelled, not as an error.
- Consider a timeout for the auto-fetch path specifically, since nobody is
  watching it.

Keep the one-credential-path rule: cancellation must not become a second auth or
network path.

## Not a duplicate of

- **#212** — tracking issue for the hardening sweep; this is the concrete missing
  capability it points at.

*From the competitor research sweep (Aug 2026).*

jonassaa:
> **Superseded by #260**, which landed on `main` first and closed #234. This is a draft kept for
> reference; the pieces with no counterpart in what shipped are tracked in #263. See the comment below.

An independent implementation of #234.

A clone, fetch, pull or push that hangs could only be escaped by force-quitting the app — our own source admitted it at `commands/create.rs:253`. All four are now stoppable, and stopping one leaves the repository exactly as it was.

Spec: `docs/superpowers/specs/2026-08-26-cancel-network-ops-spec.md` · Plan: `docs/superpowers/plans/2026-08-26-cancel-network-ops-plan.md`

## What you get

- **Clone** — while it runs, the dialog's **Cancel** button *is* the cancel (it used to be `disabled={busy}`, which is what made a stall a force-quit). It reports "Stopping…", and the dialog closes only once the backend confirms it cleaned up.
- **Fetch / pull / push** — one **Stop** button in the titlebar next to Push, shown only while a cancellable op is running and titled with what it will stop. Keyboard route: a `Stop the running <op>` palette row, gated the same way.
- **Window close** cancels everything in flight.
- **Auto-fetch** no longer stacks: it skips while a fetch is in flight, and cancels its own after two minutes.

## The decisions worth reviewing

**The registry never interrupts the op.** `cancel.rs` holds `id → Op` and kills the op's *child*; the op then notices — its `wait()` returns because git is dead, and it checks `is_cancelled()` before mapping stderr. The two shapes that read better both fail: `tokio::select!` over `child.wait()` and a cancel future needs `&mut child` in both arms, and `Arc<Mutex<Child>>` deadlocks because the op holds the lock across the very `wait()` the killer must interrupt.

**SIGTERM to the process group, not SIGKILL to the child.** Two things are wrong with `child.kill()`:

- *Wrong process.* `git` spawns `git-remote-https` (https) or `ssh` (ssh), so killing only `git` leaves the transfer's child holding the connection. `proc::git_async`/`git_async_in` now set `process_group(0)`; the `*_keeping_console` constructors deliberately do **not** (an interactive child outside the terminal's foreground group is stopped by SIGTTIN/SIGTTOU).
- *Wrong signal.* git's handlers remove its lock files (`lockfile.c`) and, in `clone`, the partial destination (`remove_junk_on_signal`). SIGKILL skips all of it, so a SIGKILLed pull strands `.git/index.lock` and the **next** pull fails with "Unable to create '…/index.lock': File exists". A cancel that breaks the following operation is worse than no cancel.

`kill_tree` verifies `getpgid(pid) == pid` before using `killpg`, falling back to a single-process kill — without that, a future spawn site that skipped the process group would have a cancel signal *our own* group. A **second** cancel escalates to SIGKILL: the user clicking again is the escalation signal, so there is no timer and no new tokio feature.

**The frontend mints the op id, before the invoke** (`src/lib/opId.ts`). A backend-assigned id emitted back would leave the op unaddressable for exactly as long as it fails to answer, which is the case this exists for. `op_id` is `Option<String>`, so `push_tag`, `push_delete_branch`, LFS, submodule update and the forge PR checkout are untouched and behave exactly as before — each is one argument from opting in.

**A cancelled clone restores the pre-clone state exactly.** `validate_clone_target` accepts only an absent target or an existing **empty** directory, so `run_clone` records which it saw: a directory *we* created is removed, a directory the **user already had** is emptied and kept. Cancel only — a failed `git clone` removes its own destination, and widening this to every failure would mean deleting a directory on a path we have not just proved is ours.

**`AppError::Cancelled` is a state, not a failure.** Checked *before* `map_git_failure` at every site, so a killed git's dying words cannot become a `Network` banner — or an `Auth` one that pops the credential dialog over a cancel. `isCancelledError` then keeps it out of every error surface, while the `refreshAll()` still happens (a partly-applied pull changed the tree).

**Windows is the abrupt case, knowingly.** No SIGTERM, and our children have no console for `GenerateConsoleCtrlEvent`, so cancel is `taskkill /F /T`. git gets no chance to clean up: `discard_partial_clone` covers the clone destination, and a stale `index.lock` after a cancelled pull is the known remaining gap, recorded in `docs/dev/backend.md`. The argv is a pure, unit-tested function so the interesting half is compiled on every platform — PR CI does not build Windows.

## Found while reviewing my own diff

Minting the op id *around* `withAuthRetry` was wrong: it resolves as soon as a credential challenge is **raised**, so the id would be cleared while the prompt was up and the retry — a second git process, every bit as able to stall — would run with no Stop button. Each attempt now mints its own id, as the clone path already did. Pinned by a test.

## Verification

- `cargo test` — 750 passing, including a **real** cancel: a TCP listener that accepts and never answers makes `git clone` hang, the test cancels it, and asserts `Cancelled` plus an absent destination. Plus the cancelled-before-spawn window and the cleanup table.
- `pnpm test` — 1950 passing (227 files), covering both surfaces, the `netOps` publish/clear pairing, the per-attempt retry id, "cancelled is not an error", and the auto-fetch skip + deadline under fake timers.
- `pnpm tsc --noEmit` and `pnpm exec tsc -p e2e/tsconfig.json --noEmit` clean.
- `pnpm test:e2e:docker full --spec create --spec remote --spec palette` — 22 passing, 3 spec files, including two new specs that cancel a genuinely stalled clone and a genuinely stalled fetch through the real IPC path:

```
[WebKitGTK 605.1.15 linux #0-0]    ✓ cancels a stalled clone, closing the dialog and leaving nothing behind
[WebKitGTK 605.1.15 linux #0-1]    ✓ Stop cancels a stalled fetch, with no error banner
Spec Files:	 3 passed, 3 total (100% completed) in 00:00:11
```

## Deliberately out of scope

Cancelling `push_tag` / `push_delete_branch` / LFS / submodule update / forge checkout (no progress surface, not the first-five-minutes hang); a timeout for *interactive* ops (a user watching a slow clone of `torvalds/linux` must not have it yanked away); resuming a cancelled clone (git has no resume — the offer would be a fresh clone, which the dialog already is).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -135,6 +135,11 @@ Each rule's full story (why, traps, tests that pin it) is in the named doc.
   exported `withAuthRetry`. Never a second auth path. Secrets travel in env,
   never argv; end option parsing with `--` before user-supplied values.
   (`docs/dev/backend.md`)
+- **One cancel path.** A long-running child is stopped through
+  `cancel::OpRegistry` — the frontend mints the `opId` before the invoke, and
+  cancel kills the child's process group with SIGTERM (never SIGKILL first: git's
+  handlers remove its lock files). A cancelled op returns `AppError::Cancelled`,
+  which no surface may render as an error. (`docs/dev/backend.md`)
 - **Forge tokens are NOT git credentials** — separate storage, separate types
   (`Secret`), no command returns a token. (`docs/dev/backend.md`)
 - **One signing chain** (`libgit2.rs::sign_payload`) for commits AND tags; a
--- a/docs/dev/architecture.md
+++ b/docs/dev/architecture.md
@@ -18,8 +18,17 @@ update.rs        Update discovery: semver compare, dev-build (0.0.0) short-circu
                  https_only. Logic only — handlers live in commands/update.rs
 proc.rs          THE only sanctioned child-process spawner (issue 172):
                  git / git_async / git_async_in (CREATE_NO_WINDOW,
-                 GIT_TERMINAL_PROMPT=0, stdin closed), program / program_async,
-                 and the two *_keeping_console exceptions. See backend.md
+                 GIT_TERMINAL_PROMPT=0, stdin closed, and — the async git two —
+                 their OWN process group so a cancel can signal the group),
+                 program / program_async, and the two *_keeping_console
+                 exceptions, which get NO process group (SIGTTIN/SIGTTOU would
+                 stop an interactive child). See backend.md
+cancel.rs        Cancelling a running clone/fetch/pull/push (#234): OpRegistry
+                 (id → Op, managed as Arc state), OpGuard (de-registers on Drop),
+                 Op::attach/cancel/is_cancelled, kill_tree (unix: SIGTERM to the
+                 process group, but killpg ONLY when getpgid(pid) == pid;
+                 windows: taskkill /F /T). The op is NOT interrupted — its child
+                 is killed and it notices via is_cancelled. See backend.md
 reveal.rs        "Reveal in Finder/Explorer" + "Open in terminal" (#215).
                  HostPlatform decouples argv building from the actual host, so
                  reveal_plan/terminal_plan are PURE and unit-tested for all
@@ -141,14 +150,18 @@ commands/        Thin Tauri handlers, one file per area:
 │                vs its first parent) and diff_commits (rev↔rev) — check arity
 ├── branches.rs  list_branches/tags/stashes/remotes,
 │                checkout/create/delete/rename_branch, set_upstream,
-│                fetch, fetch_all, pull, push,
+│                fetch, fetch_all, pull, push (all four take an optional op_id
+│                and are cancellable — #234; the rest of this file's network ops
+│                are not, and pass none),
 │                add/remove/rename/prune_remote, set_remote_url,
 │                create/delete/push_tag, verify_tag, merge_branch, rebase_onto,
 │                checkout_ref, push_delete_branch
 ├── net.rs       Shared network plumbing (Credentials, run_git_authenticated,
-│                apply_auth_env, map_git_failure, credential_approve) + ONE
-│                command, remember_credential — stored only after the credential
-│                worked. See backend.md
+│                run_git_cancellable, apply_auth_env, map_git_failure,
+│                credential_approve) + TWO commands: remember_credential —
+│                stored only after the credential worked — and cancel_operation
+│                (#234), which answers false for an id that already finished.
+│                See backend.md
 ├── update.rs    check_for_update, get_update_capability, open_url — thin
 │                handlers for src-tauri/src/update.rs (same basename, two files)
 ├── history.rs   reset, cherry_pick, revert
@@ -169,7 +182,10 @@ commands/        Thin Tauri handlers, one file per area:
 ├── lfs.rs       lfs_status, lfs_checkout (local), lfs_fetch/lfs_pull (credentialed)
 ├── bisect.rs    bisect_status/start/mark/reset
 └── create.rs    init_repo, default_init_branch, clone_repo (streaming
-                 clone://progress events)
+                 clone://progress events, optional op_id). Also
+                 discard_partial_clone (#234): on a cancel, a destination we
+                 created is removed and one that already existed is EMPTIED,
+                 never deleted
 ```
 
 ## Frontend (`src/`)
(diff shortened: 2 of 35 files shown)


jonassaa:
**Superseded by #260, which landed on `main` while this was being written.** Converted to a draft; do not merge it as-is.

#260 solves #234 with a different and reasonable design — cancellation keyed by `Scope::Clone` / `Scope::Repo(path)` at the two existing choke points, rather than by an op id threaded through the commands. Its scope keying is genuinely better in one respect I did not handle: one Cancel reaches a *pile* of stacked auto-fetches, which an id cannot. It also avoids all the command-signature churn this branch has.

Keeping this open as a draft only because a few pieces here have no counterpart in what shipped. Filed as #263 rather than argued for here:

1. **The kill is SIGKILL to the direct child.** #260's own module docs note the transport helper (`git-remote-https`, `ssh`) survives. The part not noted is lock files: git installs `remove_lock_file_on_signal` (`lockfile.c`) for SIGTERM and friends, and SIGKILL is uncatchable — so a cancel landing in the ref-update window can strand `FETCH_HEAD.lock` and fail the *next* fetch. This branch sends SIGTERM to the child's own process group (`proc::git_async*` set `process_group(0)`), with `getpgid(pid) == pid` verified before `killpg` so a future spawn site that skipped it degrades instead of signalling our own group, and escalates to SIGKILL on a second click.
2. **Window close does not cancel.** #260 relies on `kill_on_drop`, which needs the future to be dropped; process exit does not reliably unwind, so a clone can outlive the app — the case `commands/create.rs` already had a comment about. This branch cancels all in-flight ops on `CloseRequested`, gated on the `main` window so closing the `merge` resolver does not stop a fetch.
3. **No e2e coverage.** #260's `tests/net_cancel.rs` is a good Rust-level test of the same stall trick. This branch also drives both UI surfaces against the real binary in Docker (`create.e2e.ts`, `remote.e2e.ts`) using a `stalledGitRemote()` fixture.
4. **No keyboard route.** The status-bar Cancel is a mouse target; this branch adds a palette row.
5. **Auto-fetch deadline.** #260 rejected a timeout, correctly, for *interactive* ops. This branch bounds only the ops the **timer** started, where nobody is watching and a stall is indistinguishable from a hang.

Happy to close this and land 1–5 as a small follow-up on top of #260's registry instead, which is almost certainly the right shape — say the word.

jonassaa:
Closing, as offered above — #273 was "the word".

#234 is closed by #260, and this branch's design (an op id threaded through the
commands) was superseded on merit by #260's scope keying, which handles the
stacked-auto-fetch case an id cannot. The one genuinely load-bearing piece here,
SIGTERM to the child's own process group before SIGKILL, has been re-landed on
top of #260's registry as #273 — the right shape, and without this branch's
command-signature churn.

What is left of this branch is items 2-5 of #263, each of which wants a small PR
against `main` rather than a rebase of 38-commits-behind work built on a design
`main` did not take.

**The branch is not deleted.** `feat/cancel-net-ops` @ e615c1b stays fetchable,
so the reference implementations #263 points at are still readable at their
paths on it:

- `e2e/support/tempRepo.ts` — `stalledGitRemote()`, plus the cancel specs in
  `e2e/specs/create.e2e.ts` and `e2e/specs/remote.e2e.ts` (#263 item 3)
- `src-tauri/src/lib.rs` — `CloseRequested` cancelling in-flight ops, gated on
  the `main` window (item 2)
- `src/features/palette/commands.ts` — the keyboard route (item 4)
- `src/AppShell.tsx` — the deadline on timer-started fetches (item 5)

Read them for the shape, not the plumbing: anything lifted needs re-keying to
`Scope` rather than an `op_id`.

PR #280, opened 2026-08-27
fix(distribution): the publisher cannot be the product name either

jonassaa:
`bundle.publisher` had **two** wrong values available. #278 fixed the first (`github`, the identifier fallback). This is the second.

`"platypusgit"` equals `productName`, and Tauri's [Microsoft Store guide](https://v2.tauri.app/distribute/microsoft-store/) states: *"Your application publisher name cannot match the application product name."* It reads correct, which is exactly what makes it worth a guard rather than a comment.

## Why now, specifically

The window is open and closing:

- **No published release carries it.** v0.1.1 predates #278, so today this is one line and nobody's registry moves.
- **After v0.1.2** it becomes a second registry-identity change for everyone who installed.
- **After the first winget submission** it is locked in by `AppsAndFeaturesEntries.Publisher`, which has to match what the installer actually wrote.

`productName` is the field that can't move — the bundler derives the Debian package name from it (#269) — so the publisher is what changes. `"Jonas Aasberg"` is also just the accurate answer: the publisher is who ships it, not what is shipped.

## Why a fourth test

The three existing guards are deliberately value-agnostic — non-empty, and `!=` the identifier fallback — so **all three pass with either value**. Nothing in the suite could see this. Shown failing on the bad input, not just passing on the good one:

```
test the_publisher_is_not_the_product_name ... FAILED
  assertion `left != right` failed: bundle.publisher equals productName
  (`platypusgit`). The Microsoft Store rejects that pairing, so this blocks
  the MSIX channel.
test the_bundle_names_a_publisher ... ok
test the_publisher_is_not_the_identifier_fallback ... ok
test the_wix_upgrade_code_is_pinned ... ok
```

## Blast radius

`UpgradeCode` untouched → in-place upgrade unaffected. The HKCU shortcut-bookkeeping key moves from `Software\platypusgit\platypusgit` to `Software\Jonas Aasberg\platypusgit`; it holds shortcut flags and a `PrevInstallDir` search, the install location is unchanged, so an upgrade lands in the same directory. Not marked `!` because the value it replaces has never shipped.

## Verification

- `cargo test --test msi_identity` → 4 passed; negative case confirmed above.

Config + test only; nothing reaches the e2e snapshot.

Context: `docs/superpowers/specs/2026-08-27-msix-store-spec.md` §F (landing separately on `feat/msix-store`).

### DIFF
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -41,7 +41,7 @@
     "createUpdaterArtifacts": true,
     "targets": ["msi", "dmg", "deb", "appimage"],
     "category": "DeveloperTool",
-    "publisher": "platypusgit",
+    "publisher": "Jonas Aasberg",
     "shortDescription": "A developer-focused git client",
     "longDescription": "PlatypusGit — a cross-platform, developer-focused git client.",
     "copyright": "Copyright (c) 2026 Jonas Aasberg",
--- a/src-tauri/tests/msi_identity.rs
+++ b/src-tauri/tests/msi_identity.rs
@@ -74,6 +74,32 @@ fn the_publisher_is_not_the_identifier_fallback() {
 }
 
 #[test]
+fn the_publisher_is_not_the_product_name() {
+    let conf = conf();
+    let product = conf["productName"].as_str().expect("productName is a string");
+    let publisher = conf["bundle"]["publisher"]
+        .as_str()
+        .expect("bundle.publisher is a string");
+
+    // A THIRD wrong value for this field, and the only one that looks right.
+    // Tauri's Microsoft Store guidance: "Your application publisher name cannot
+    // match the application product name." `productName` is lowercase and
+    // load-bearing for the Debian package name, so the field that has to move is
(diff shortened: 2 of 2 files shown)


jonassaa:
Heads-up: this landed independently as #279 about half an hour ago — same
diagnosis, same value (`Jonas Aasberg`), same file. We were in flight at the
same time and neither saw the other, which is the second time today this pair of
findings has been discovered twice (#277 documented them, #278 fixed the first
half).

`main` already has the config change, byte-identical, so the only real delta is
the guard:

- **#279's version is case-insensitive** (`publisher.to_lowercase() != product.to_lowercase()`).
  That was deliberate — `PlatypusGit` vs `platypusgit` is the variant someone
  reaches for next when this test fails, and the Store constraint is about the
  displayed names rather than the bytes. It is shown failing on exactly that
  near-miss in the PR body.
- This PR's `assert_ne!(publisher, product)` would pass on `PlatypusGit`.

So closing this as superseded loses nothing — unless you want the spec
cross-reference in its doc comment, in which case that is worth a one-line
follow-up pointing `msi_identity.rs` at
`2026-08-27-msix-store-spec.md` §F once #281 merges. Happy to do that if useful.

Issue #212, opened 2026-08-25
Stability pass: clone, stage/commit, push/pull and diff must survive launch day


jonassaa:
Before any launch push, the four flows that every single new user touches in
their first five minutes need to be hardened deliberately. If clone,
stage/commit, push/pull or diff viewing breaks under an unfamiliar repository,
that user never opens the app again — and says so publicly.

This is a tracking issue for a focused hardening sweep. **Every failure found
gets its own issue, linked back here.** The point is to go looking for the
breakage on purpose instead of finding out from a stranger.

## The four flows

- **Clone** — `src-tauri/src/commands/create.rs` (`run_clone`, `clone_repo`,
  `validate_clone_target`), `src/features/create/CloneDialog.tsx`.
  e2e: `e2e/specs/create.e2e.ts`
- **Stage / commit** — `src/screens/CommitPanel.tsx`, `commands/commits.rs`.
  e2e: `e2e/specs/status-stage.e2e.ts`, `commit.e2e.ts`
- **Push / pull** — `commands/net.rs` (`run_git_authenticated`),
  `useRepoStore`'s `withAuthRetry`. e2e: `e2e/specs/remote.e2e.ts`
- **Diff viewing** — `src/features/diff/`, `src/lib/diffRows.ts`,
  `commands/diff.rs`. e2e: `e2e/specs/diff-nav.e2e.ts`,
  `diff-selection.e2e.ts`, `history-diff.e2e.ts`

All four already have e2e coverage on the happy path. This sweep is about the
unhappy paths, which is where a first-run user actually lands.

## Adverse conditions to walk each flow through

Not a code review — actually drive the app.

**Clone**
- [ ] URL that does not exist / 404 / typo'd host — is the error legible, and is
      the dialog still usable afterwards?
- [ ] Private repo, wrong credentials, then cancelled credential prompt
- [ ] Target directory already exists / is non-empty / is not writable
- [ ] A genuinely large repo (try `torvalds/linux`) — does progress report, can
      it be cancelled, does the UI stay responsive?
- [ ] Network drops mid-clone
- [ ] Repo with submodules, with and without recursion
- [ ] Empty repository (no commits) — the `Unborn` path

**Stage / commit**
- [ ] Empty commit message; whitespace-only message
- [ ] No user.name / user.email configured at all — a fresh machine has none.
      Is the error actionable or does it just say "failed"?
- [ ] Untracked file discard (regression watch: #67)
[...]

PR #424, opened 2026-09-07
fix(diff): say so when an image preview will not decode

jadhavgaurav:
Finding 8 from the code audit in #212: *"A corrupt or undecodable image renders a broken `` with no fallback."*

## The problem

`image.rs` sniffs the first 512 bytes, so a truncated or corrupt image whose magic number is intact comes back as `kind: "image"`. `ImageDiffView` then hands those bytes to an `` that has an `onLoad` handler and no `onError`. When the webview refuses to decode them, the panel shows the browser's broken-image glyph and says nothing, which is exactly what `image.rs`'s module doc rules out: *"a broken `` is worse than a sentence."*

Every other non-previewable state here already says something specific (too large, LFS not fetched, SVG refused). This was the one that said nothing.

## The change

- `onError` on the preview `` retires that side to the same note treatment the other states use: "This image could not be decoded. The file may be truncated or corrupt."
- The byte caption stays. It came from the backend, not from the decoder, so it is still true.
- The failed set is per side and is keyed and cleared exactly like `dims`, so a corrupt new version still shows the readable old one, and the next selection starts clean rather than printing the previous file's failure over a good image.

All four diff surfaces inherit it, since they all render this one component.

## Tests

Three cases in `ImageDiffView.test.tsx`, driving the real `error` event: the sentence replaces the image and keeps the caption; only the failing side fails; a new path clears the failure. All three fail on `main` and pass with the change (verified by reverting the component and re-running).

## Checks

- `pnpm vitest run` — 355 files / 3542 tests, all passing (includes the `docs` project)
- `pnpm tsc --noEmit` — clean
- `pnpm vite build` — clean
- `cargo check --manifest-path src-tauri/Cargo.toml` — clean
- `cargo test --manifest-path src-tauri/Cargo.toml` — all green (no Rust changed; run to confirm nothing moved)

`docs/dev/frontend.md` gains a bullet under the `ImageDiffView` section describing the new state.

No e2e spec was run: there is no fixture with a deliberately corrupt image, and adding one to drive a decode failure through the real webview would be a larger change than the fix. Happy to add it if you would rather have it covered there.

### DIFF
--- a/docs/dev/backend.md
+++ b/docs/dev/backend.md
@@ -480,6 +480,27 @@ escape hatch from becoming the uncapped path #385 removed.
   `\x89PNG` is `NotAnImage`, `BM` needs its reserved field zero, an ICONDIR
   needs a non-zero image count. A repository is untrusted input, and a broken
   `<img>` is worse than a sentence.
+- **A sniff is not a validation — `integrity` is, and it has to live here.** The
+  header of a file whose data was cut off is intact, so `sniff` answers `Image`
+  for an aborted download, a bad merge or a half-written `git lfs smudge`. **A
+  webview does not report that as an error.** Measured in Blink 152: a PNG cut
+  to 60% of its length, and one whose IDAT was overwritten at the same total
+  length, both fire `load`, both report `naturalWidth/Height` from the intact
+  header, and one paints 60% of the picture while the other paints nothing —
+  `error` fires only when there is no decodable header at all. So no frontend
+  signal can catch this class, and `git/image.rs::integrity` — which holds the
+  WHOLE blob, the only place that does — refuses it as
+  `UnsupportedReason::Truncated` before it can reach an `<img>`.
+  **It reports only what the bytes PROVE**, because a false "truncated" hides an
+  image that would have displayed: PNG's chunk chain must reach a whole `IEND`;
+  a JPEG with no `FF D9` anywhere is proven short (that marker cannot occur in
+  stuffed scan data — though its presence proves nothing, since an EXIF
+  thumbnail carries one, so a photo cut after its thumbnail reads as complete);
+  WebP, BMP and ICO each declare their own length. CRCs are NOT verified
+  (decoders in the wild ignore them) and **GIF is deliberately not checked at
+  all** — one unframed `0x3B` trailer byte cannot prove anything. All of it is
+  pinned in `image.rs`'s tests, including a sweep that cuts a real encoder's PNG
+  at every offset and asserts each one still sniffs as an image.
 - **SVG is recognised and REFUSED**, as its own `UnsupportedReason::Svg` so the
   UI can say so rather than looking broken. It is the one format on the list
   that is not inert (script, `<foreignObject>`, external `href`/`url()`,
--- a/docs/dev/frontend.md
+++ b/docs/dev/frontend.md
@@ -152,6 +152,14 @@ Part of the `docs/dev/` set (`architecture`, `testing`, `frontend`, `backend`,
     the webview already knows, and until it fires there simply are no
     dimensions (never `0 × 0`). The backdrop is a checkerboard built from
     `--bg-1`/`--bg-2` so transparency reads in both themes (#236).
+  * **A side the webview refuses to decode says so** (#212). `image.rs` sniffs a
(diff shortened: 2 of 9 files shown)


jonassaa:
Reviewed this, and pushed a follow-up commit onto the branch rather than asking for a round trip. Your commit is untouched — it was a fast-forward, not a force-push.

**Your diagnosis and your fix are both right, but `onError` catches far less than the wording claims.** I built PNG variants and measured the real events in Blink 152 (which is exactly WebView2's engine):

| case | event | naturalSize | what renders |
|---|---|---|---|
| valid (control) | `load` | 64×64 | the image |
| **truncated to 60%** (cut mid-IDAT) | **`load`** | 64×64 | top 60%, rest blank |
| truncated to 90% | `load` | 64×64 | looks fine |
| **corrupt IDAT** (garbage body, same length) | **`load`** | 64×64 | **completely blank** |
| magic bytes only (8 bytes) | `error` | 0×0 | broken glyph |
| magic + IHDR only | `error` | 0×0 | broken glyph |

So the two cases the sentence names — truncated and corrupt — are precisely the two an `` cannot report. It decodes the prefix, fires `load`, reports the intact header's dimensions and paints what it got, so the panel captions a half-blank picture `64 × 64` and says nothing. `error` only fires when there is no decodable header at all. The corrupt-IDAT row is the worst of them: a totally blank panel claiming to be a 64×64 image, which reads as a legitimately transparent PNG.

**What the follow-up commit adds**

- `git/image.rs::integrity` decides truncation from the WHOLE blob — the only place that holds one. PNG's chunk chain must reach a complete `IEND`; a JPEG with no `FF D9` anywhere is proven short (that marker cannot occur in stuffed scan data); WebP, BMP and ICO each declare their own length. It becomes `UnsupportedReason::Truncated` and never reaches an ``.
- It reports **only what the bytes prove**. A false "truncated" would hide an image that displays, which is worse than the broken preview it replaces — so CRCs go unverified (decoders in the wild ignore them) and GIF is not checked at all, because one unframed `0x3B` trailer byte proves nothing.
- Your `onError` stays, as the backstop for the one thing only the engine knows: a structurally complete file this webview will not display. Its sentence now names no cause, since the app ships on three engines whose format coverage differs — "truncated or corrupt" would be a lie on the machines where the same file opens fine.

**Two things I changed in your code, with reasons**

1. `dims` and the decode failure are now keyed to the **bytes** they describe rather than cleared by an effect. The reset keyed on `[path, previewKey]`, which was narrower than the identity `useImagePreviews` actually reads on — it ignores `repoId` and each side's `source`, and stands in `data.length` for the data. Same path at a different revision with an equal byte length kept the previous blob's state: a caption for the wrong image, or a failure notice over a good one. Deriving it also can't race the browser's `error` task against a passive effect, and one side's new bytes can no longer clear the other side's live failure.
2. `PanelNote.icon` is typed `IconName` instead of `string`. I checked this by planting `icon: "wanr"` — it passed **both** `tsc` and `test/iconSet.test.ts`, because that guard only scans `name="…"`/`icon="…"` attributes and never object literals. It's now a hard type error. That gap pre-dated your commit; your `UNDECODABLE` was just the fourth literal to fall through it.

**On the tests:** all three of yours did fail on `main` and pass with your change — I verified that by reverting the component and re-running. Two of them needed rewriting once state is keyed to bytes, though, because they answered both paths with the *same* base64 (`png()` defaults to `data: "AAAA"`). Identical bytes are the same image, so "a perfectly good image" was byte-identical to the one that had just failed, and only jsdom's refusal to decode anything hid it. They use genuinely different blobs now. I also added the case the old reset never had: same path, different revision, equal byte length — the `previewKey` half of that dep array could be deleted outright without failing a single test, which I confirmed before changing it.

Gates on the branch: `cargo test` green (91 suites), `pnpm test` 355 files / 3545 tests, `tsc --noEmit` and `vite build` clean. Both new guards were proven by planting their violations.

Note this does **not** fully close #212's finding 8 — a JPEG truncated after its EXIF thumbnail still reads as complete, deliberately, because under-reporting is the safe direction. Worth leaving the audit item open with that caveat.

Thanks for this — the diagnosis was sound and the `` really was the only place saying nothing.

PR #466, opened 2026-09-14
chore(deps): September Dependabot batch, plus the smol-toml advisory

jonassaa:
Folds the six open Dependabot PRs into one branch, declines one of them with a
reason, and closes the second-to-last open security advisory while the lockfile
is already being regenerated.

Bundling is the measured-cheaper shape here: `pnpm-lock.yaml` peer fanout makes
every npm PR conflict the rest on each merge, so merging them serially costs a
rebase and a full CI run apiece. One branch, one CI run, no rebases.

### Taken

| PR | Change |
|----|--------|
| #460 | `react`, `react-dom`, `@types/react`, `@types/react-dom` 19.2.x → 19.3.0 |
| #462 | `lucide-react` 1.41.0 → 1.45.0 |
| #463 | `@wdio/{cli,local-runner,mocha-framework}` → 9.31.9, `@wdio/tauri-service` 1.3.0 → 1.4.0 |
| #465 | `vite` 7.3.6 → 8.3.0 |
| #461 | crates `tauri-plugin-wdio-webdriver` 1.3.0 → 1.4.0, `uuid` 1.26.0 → 1.26.1 |

### Declined

**#464 — `@types/node` ^22 → ^26.** `@types/node` describes the *runtime*, and
every workflow here pins `node-version: 22`, CLAUDE.md pins the toolchain to
Node 22, and `Dockerfile.e2e` tracks 22.x. Types for Node 26 would let code
type-check against APIs that exist nowhere this app is built, tested or
shipped — and `tsc` would be green about it. The specifier has read `^22` since
it was introduced in #26 and has never been bumped, so this records the standing
decision as a `dependabot.yml` ignore rule scoped to semver-major (22.x patches
still flow) and gives back one of the five weekly PR slots.

### Security fix

**Dependabot alert 67 — `smol-toml`, GHSA-7w5x-hrqm-74c2** (high, dev scope):
DoS from malformed TOML documents. This is one of the two advisories
`docs/dev/testing.md` recorded as knowingly left open after the extract-zip
prune. It was left alone because the vulnerable range was `<= 1.7.0` — every
published version, so there was no number to floor to. 1.7.1 shipped the fix.

An override is the only route: it arrives via `@wdio/native-utils`, which the
block pins *exactly* to 2.5.0 to escape the broken 2.4.0, so it cannot be bumped
out from the parent. `@wdio/native-utils` declares `smol-toml: ^1.6.0`, so
`^1.7.1` resolves inside its own range and lands on 1.8.0 — nothing is forced
across a major. Guarded as an ordinary floor + required key in
`test/depOverrides.test.ts`, with the reasoning in `docs/dev/testing.md`.

That leaves `glib` as the only open alert, which ships, is documented in
`docs/dev/distribution.md`, and is still not ours to fix — so the triage tracked
in issue 346 has no actionable items left after this.

### `vite` 8 — why it is safe to take

- `@vitejs/plugin-react` 5.2.0 already declares `vite: ^8.0.0`, so the React
  plugin needs no companion bump.
- vite 8 keeps the same `esbuild` range (`^0.27.0 || ^0.28.0`), so the
  `esbuild` 0.28.1 floor — the one FLOORS entry with no override key — still
  holds.
- It swaps rollup for rolldown. Production bundle builds clean; the entry chunk
  is unchanged at 1,612.30 kB.
- `vite.config.ts` moves `__dirname` → `import.meta.dirname` (vite 8 warns the
  native config loader will not support it), and the worker comment stopped
  naming rollup since it is no longer the bundler refusing an `iife` split
  build.

### Lockfile audit

Specifiers were edited by hand and the lockfile regenerated from `main`'s, never
from Dependabot's, so none of its unrelated churn came along. After
regeneration: the 15-key `pnpm.overrides` block is intact, `esbuild` stays at
0.28.1, `extract-zip` still resolves nowhere, `@types/node` and
`@tauri-apps/api` did not move, and the `importers:` diff touches only the
specifiers listed above.

### Verification

- `pnpm test` — **403 files, 4141 tests pass** (up 2: the new `smol-toml` floor
  and key assertions).
- `cargo test --locked` — **1334 pass, 0 fail**.
- `pnpm tsc --noEmit` and the e2e typecheck both clean.
- `pnpm vite build` clean under rolldown.
- Guard proven by planting the violation, and the two shapes differ in a way
  worth recording: deleting the override key but leaving the lockfile gives
  **1** failure (pnpm keeps the resolved 1.8.0, so the floor passes
  *vacuously*); deleting the key and regenerating from `main` — the shape an
  actual Dependabot PR arrives in — gives **2**.
- e2e is left to the required `e2e-linux` check, which runs the full suite on
  the same Docker stack. `docs/dev/testing.md` is explicit that an overrides
  change is not proven by `pnpm test`, and these bumps touch the runner itself.


jonassaa:
One more measurement worth recording, taken on this branch after the last edit:

```
$ pnpm audit --prod
No known vulnerabilities found

$ pnpm audit          # ALL scopes, dev included
No known vulnerabilities found
```

`docs/dev/testing.md` last recorded this as "`pnpm audit` goes from 4 advisories
to 2, both remaining ones (`deepmerge-ts`, `smol-toml`) pre-existing and
untouched". With `smol-toml` overridden here and `deepmerge-ts` since resolved,
**the npm side is now at zero on every scope**.

`glib` remains the only open Dependabot alert repo-wide. It is a Rust runtime
advisory, is documented in `docs/dev/distribution.md`, and is still blocked
upstream — not something this PR can or should touch.

Issue #346, opened 2026-09-01
Dependabot triage: 24 open alerts, none of which Dependabot can close


jonassaa:
## Bottom line

24 Dependabot alerts are open. **None of them reach a shipped binary via npm** — all 22
npm alerts are `development` scope, and none of the affected packages appear in
`dependencies`. The two Rust alerts are `runtime` scope; one is build-time-only in
practice, the other is a real Linux runtime dep that we cannot fix ourselves.

The actionable finding is not any single CVE — it is that **Dependabot cannot close a
single one of these**, so they will sit open indefinitely until someone writes a
`pnpm.overrides` block by hand.

| | count |
|---|---|
| total alerts ever | 63 |
| fixed | 33 |
| auto-dismissed | 6 |
| **open** | **24** (22 npm dev-scope, 2 rust runtime-scope) |

## Why nothing has happened automatically

Dependabot security updates are **enabled and not paused**
(`/repos/:owner/:repo/automated-security-fixes` → `{"enabled":true,"paused":false}`),
yet there are **zero open Dependabot PRs**.

Every open npm alert is **transitive-only** — the vulnerable package is never a direct
dependency of ours. Dependabot's security updater can only bump a manifest entry; it
does not write `pnpm.overrides`. So for each of these it opens nothing and the alert
stays open forever.

This also explains the asymmetry with `site/`: every `site/**` alert is `fixed`,
because those were direct-dependency bumps (`astro`, `sharp`, `svgo`, `postcss`) that
the monthly grouped PR swept up. The root project's alerts are all transitive, so
nothing sweeps them.

Related: **there is no `pnpm audit` or `cargo audit` step in any workflow**
(`.github/workflows/{tests,e2e,release,site}.yml`). The GitHub security tab is
currently the only thing that knows about any of this.

## What ships and what does not

The shipped frontend bundle is built from `dependencies` only — react, react-dom,
codemirror, shiki, zustand, `@tauri-apps/*`, lucide-react, node-diff3, fontsource.
**Not one affected package is in that set.** Every affected package is reachable only
from `devDependencies`, and the blast radius is a developer machine or a CI runner,
not a user install.

Notably, **`@wdio/utils` → `@puppeteer/browsers` is the single worst subtree**: it
pulls `extract-zip` (the one unpatchable high) *and* the whole `proxy-agent` →
`socks-proxy-agent` → `socks` → `ip-address` chain (3 more alerts). That is 4 of the
[...]

PR #469, opened 2026-09-14
ci: scheduled dependency audit, and the two high-severity advisories it found

jonassaa:
Closes the last actionable item on #346 — and the job found two high-severity
advisories the moment it was pointed at this tree.

### The advisories (first commit)

`cargo audit` had never been run against this repo. The first run exits **1**
with two **7.5 high** advisories against `quick-xml` 0.38.4:

| | |
|---|---|
| RUSTSEC-2026-0194 | Quadratic run time checking a start tag for duplicate attribute names |
| RUSTSEC-2026-0195 | Unbounded namespace-declaration allocation in `NsReader` — memory-exhaustion DoS |

**Neither had a Dependabot alert.** GitHub's advisory database and the RustSec
database do not agree, and this is what fell through the gap — worth
remembering the next time "no open alerts" gets read as "nothing to fix".

The chain is `quick-xml` ← `plist` ← `tauri`, so it reaches the shipped binary.
No manifest edit is needed: `tauri`'s range already admits a newer `plist`, so
`cargo update -p plist` moves 1.8.0 → 1.10.1 and carries `quick-xml` to 0.42.0,
past the `>= 0.41.0` the advisories ask for. **Lockfile only.**

### The job (second commit)

`#346`'s standing finding was never a single CVE — it was that 24 alerts had
piled up where nothing could act on them. Dependabot's security updater only
bumps a *manifest* entry, and every npm advisory here is transitive-only, so it
opens nothing and the alert sits in the security tab forever. No workflow had
ever run an audit, so that tab was the only thing that knew.

**It is deliberately not a gate.** No `pull_request` trigger at all — the only
reliable way to stop a reporting job quietly becoming one. A new advisory in a
dev-only transitive dep must never turn `main` red or block an unrelated PR;
`e2e-linux` stays the one required check. It runs weekly on Thursday, offset
from Dependabot's Monday, plus `workflow_dispatch`.

What *does* fail is narrow: on npm only `pnpm audit --prod` gates, since
`dependencies` is what the shipped bundle is built from, and the all-scopes run
reports into the job summary; on cargo the default gates, because those crates
ship.

**No `--ignore` list**, which is a measurement and not an omission. `cargo
audit` fails on *vulnerabilities* and merely warns on `unmaintained`/`unsound`,
so the 11 warnings here pass on their own — `glib` (RUSTSEC-2024-0429), the
shipped upstream-blocked one, among them. It needs no waiver because it never
fails the step. That is also why the step must never grow `--deny warnings`:
it would make the job permanently red over advisories nobody can act on and
train everyone to ignore it, which is the exact failure this workflow exists to
end. Written into `docs/dev/testing.md` so it is not re-litigated.

`site/` is out of scope on purpose — its dependencies are *direct*, so
Dependabot's monthly grouped PR fixes them itself.

### Verification

- `cargo audit`: **exit 1, "2 vulnerabilities found"** → **exit 0, 11 warnings**.
- `cargo test --locked`: **1334 passed, 0 failed** — unchanged from before the
  bump, so the `plist` move is behaviour-neutral.
- `pnpm test`: **403 files, 4141 tests pass**.
- **The workflow was actually run, not just linted.** `workflow_dispatch` needs
  the file on `main`, so it was proven the way `release.yml` changes are proven
  here: a temporary `push:` trigger on this branch, observed green, then dropped
  before merge. Both jobs pass — see the run linked in the comments.


jonassaa:
**Proof the workflow actually runs**, since `workflow_dispatch` cannot reach a
file that is not yet on `main`: a temporary `push:` trigger was added on this
branch, the run observed, and the trigger then dropped (the branch was
force-pushed, so that commit is no longer in this PR — the file here has
`schedule` + `workflow_dispatch` only).

Run: https://github.com/jonassaa/platypusgit/actions/runs/34837045902 — both
jobs green.

Two lines from its log confirm the steps did real work rather than no-opping:

```
npm    Production scope (gating)  No known vulnerabilities found
cargo  cargo audit               Scanning Cargo.lock for vulnerabilities (643 crate dependencies)
cargo  cargo audit               warning: 11 allowed warnings found
```

Note what the second pair shows: `cargo audit` finishes on **11 warnings and no
vulnerability error**, which is exactly the behaviour the "no `--ignore` list"
comment in the workflow describes — and it is green *because* the `quick-xml`
fix is in the first commit. Without it the same step exits 1 with
`error: 2 vulnerabilities found!`.

